PR #13, opened 2019-02-02
Add local GNU Makefile and Windows nmake Makefile for kyber768

dstebila:
This PR adds a relatively simple GNU Makefile for crypto\_kem/kyber768.  While not strictly necessary for our build system, this shows someone how it could be built without them having to figure out what's going on in our top-level Makefile.

Tested on macOS 10.14.2 and Ubuntu 18.04.

(PR #14 shows what it's like using CMake.  At most one of the two approaches should be taken.)


cryptojedi:
> @dstebila requested your review on: PQClean/PQClean#13 Add local GNU Makefile for kyber768.

I wouldn't link in nonrandombytes. That's calling for trouble if anybody
ends up really using this. Could just leave randombytes() as a
dependency that needs to be implemented outside somewhere.

dstebila:
Okay, but then we can't make a shared library, because randombytes will be an undefined symbol.

cryptojedi:
> Okay, but then we can't make a shared library, because randombytes will be an undefined symbol.

For illustration purposes it should be ok to compile to a .a, right?

dstebila:
> > Okay, but then we can't make a shared library, because randombytes will be an undefined symbol.

> For illustration purposes it should be ok to compile to a .a, right?

We could do that. I suppose the nice thing about a shared library is the compiler tells you if there are symbols missing.  But our integration tests would be catching that, I guess.

cryptojedi:
> > > Okay, but then we can't make a shared library, because randombytes will be an undefined symbol.
> 
> > For illustration purposes it should be ok to compile to a .a, right?
> 
> We could do that. I suppose the nice thing about a shared library is
> the compiler tells you if there are symbols missing.  But our
> integration tests would be catching that, I guess.

Yes, they would. The alternative would be to link in a full-fledged
portable implementation of randombytes. The code in libsodium might be a
good starting point. However, I think this is taking it a bit too far if
the main reason to include it is to make compilation to an .so work in a
Makefile that is only there to illustrate how to compile the code.

thomwiggers:
So the main reason to have these Makefiles appears to be illustrative purposes. And we're going to be complicating this example by linking as a static library, because we don't want to include a proper `randombytes`...

I would vote for having _no_ Makefiles in folders, because it:

* Suggests that the build works by calling out to these Makefiles, giving the wrong impression to people coming to PQClean
* Suggests that implementers can put their own options in there and circumvent our build options (e.g. add `-lssl`). They'd be told wrong by our CI, but it would give the wrong impression.

We could instead serve the illustrative purpose by just writing a few lines into the `README.md`:

> ### Building schemes
> 
> To build one of the schemes outside of this repo, just grab the files from `common` and build as follows:
>
>     schemename.so: scheme/clean/*.c common/*.c
>           $(CC) $(CFLAGS) -share -o  $@ $^

dstebila:
@cryptojedi Ready for review

dstebila:
I assume the Travis failures are related to the build failure on master HEAD, and not related to my changes here.

thomwiggers:
Sorry for breaking the build; I should've waited two minutes before going home after pushing :sweat_smile:

thomwiggers:
I have rebased the branch; that should fix the build.

dstebila:
Thanks Thom! Builds are working so it's ready for review and merge.

crypto_kem/kyber768/clean/GNUmakefile
```diff
@@ -0,0 +1,15 @@
+LIB=libkyber768_clean.a
```
thomwiggers: This file is named `GNUmakefile`. Are we using any GNU-specific things, or should this just be `Makefile`? There are platforms that won't necessarily have GNU Make (BSD comes to mind).

crypto_kem/kyber768/clean/indcpa.c
```diff
@@ -133,7 +133,7 @@ void PQCLEAN_KYBER768_gen_matrix(polyvec *a, const unsigned char *seed,
     uint16_t val;
     unsigned int nblocks;
     const unsigned int maxnblocks = 4;
-    uint8_t buf[SHAKE128_RATE * maxnblocks];
+    uint8_t buf[SHAKE128_RATE * /* maxnblocks = */ 4];
```
thomwiggers: why this change?

scripts_windows/build_all.bat
```diff
@@ -0,0 +1,19 @@
+@ECHO OFF
+SETLOCAL
+SET EL=0
```
thomwiggers: What does this variable do?

crypto_kem/kyber768/clean/GNUmakefile
dstebila: I am not sure, I guess I could install a BSD VM to test.  Is BSD a targetted platform?

The reason I called it GNUmakefile is that apparently GNU Make will search filenames in the following order: GNUmakefile, makefile, Makefile.  Whereas Windows nmake will search for MAKEFILE (case insensitive) and then give up.  If I called this one "Makefile", then Windows nmake would see it and fail, whereas if I call it GNUmakefile, then GNU Make will still see it, but Windows nmake won't.

crypto_kem/kyber768/clean/indcpa.c
dstebila: Microsoft compiler fails, complaining that arrays need to be initialized using constant size values.  (Yes, I'm aware that maxnblocks was declared const, but Microsoft's compiler is apparently not aware.)

scripts_windows/build_all.bat
dstebila: `EL` is a local variable to store the overall "errorlevel" (return value) that we want to return from the script.  Notice that later lines set EL to non-zero when a program that the batchfile calls itself returns a non-zero errorlevel.  (My searching seemed to indicate there is no way to automatically propagate non-zero errorlevels up like in bash's `set -e`.)

crypto_kem/kyber768/clean/GNUmakefile
thomwiggers: OS X is a BSD flavour, though I'm also just making assumptions here. It seems like lowhanging fruit to be more inclusive.

The Windows thing may be a  good reason to name it `GNUMakefile`, or we could just rely on nmake respecting `/f` (which I assume makes it skip this search) and accept that the (probably a minority) of Windows devs will just have to specify it (or throw away the Unix Makefile).

dstebila: Or I could just install BSD Make on my Mac.  Which I have done.  bsdmake fails with our GNUmakefile: the following two lines don't work:

    SOURCES=$(wildcard *.c)
    OBJECTS=$(patsubst %.c,%.o,$(SOURCES))

If I explicitly list the .o files in the makefile (like I had to do for Makefile.Microsoft_nmake) then we could have the same makefile to work for both GNU Make and BSD Make.

thomwiggers: Ah, I had no idea that these were GNU extensions. Seems too annoying to work without them though, so maybe we can also trust that BSD people know how to solve this problem :man_shrugging:.

dstebila: Well actually it's not that annoying to just list the .o files by name, and then it becomes an even more obvious makefile, so I've made that change, and then it's good for both GNU and BSD Make, and probably other random Makes out there.  Once Travis CI passes I'll merge it.

PR #14, opened 2019-02-02
Add local CMake configuration for kyber768

dstebila:
This PR adds a CMake configuration for crypto\_kem/kyber768.  While not strictly necessary for our build system, this shows someone how it could be built without them having to figure out what's going on in our top-level Makefile.

The user would have to have CMake installed to generate the Makefile.  But one advantage is that CMake can also generate Visual Studio solution files so we can have this build on Windows.

Tested on macOS 10.14.2 and Ubuntu 18.04 **and Windows 10**.

(PR #13 shows what it's like using a simple GNU Makefile.  At most one of the two approaches should be taken.)


crypto_kem/kyber768/clean/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+cmake_minimum_required(VERSION 3.0)
+
+if(MSVC)
+    add_compile_options(/Wall /Wx)
+    # /Wall causes lots of errors, weaker options are /W1 or /W2 or /W3 or /W4
```
dstebila: /Wall causes lots of errors for kyber768, even /W1 catches a few things, so we'd have to do some more cleaning in the kyber768 code.

PR #17, opened 2019-02-08
Add makefiles for Dilithium-iii

dstebila:


dstebila:
@cryptojedi The dilithium-iii/clean build fails on Windows because Microsoft's C compiler doesn't handle the `__attribute__((aligned(32)))` on line 10 of poly.h.  I don't fully understand aligned stuff in C so I'm not sure how to fix that.

Issue #10, opened 2019-01-23
Run tests from a more sophisticated script


thomwiggers:
The `Makefile` is getting unwieldy, and it turns out that some tests can't be run on all platforms.

For example, Valgrind is broken on MacOS. This makes the build even more annoying, as the `run-valgrind` target is now conditional.

We'd also like to produce all errors in one go; not abort after the first make target failed.

It's probably better then to write a Python or more advanced shell script to make these decisions and filter for us. It would probably also help reduce the output from Make (which produces a LOT of noise).

dstebila:
I think it's a good idea to move most of the checks out of the top-level Makefile and into bash or Python scripts.

Whatever we do here, we want to have sufficiently relevant messages logged to stdout, because this is all that you can see when a Travis build fails.

dstebila:
Looking at it more closely, I see you've started that with `check_*.py` scripts in the `test` directory, some of which have the semantics that you provide a scheme name `crypto_kem/kyber768` as the input.  But there are still Makefile targets for `run-testvectors-all`, etc.  

Could we have a single Makefile target `run-checks` which then basically does: 

    for all scripts in test/check*.py
        for all schemes in (crypto_kem/*, crypto_sig/*):
            $script $scheme

We could even wrap this in its own script which nicely catches errors and organizes log output.

Benefits: the Makefile stays short.  

Drawbacks: harder for a developer to run all of a certain type of test (e.g., just run-testvectors-all).

cryptojedi:
Douglas Stebila <notifications@github.com> wrote:
> I think it's a good idea to move most of the checks out of the
> top-level Makefile and into bash or Python scripts.
> 
> Whatever we do here, we want to have sufficiently relevant messages
> logged to stdout, because this is all that you can see when a Travis
> build fails.

I fully agree. Let's aim for one script per test.

dstebila:
I have some time today and could work on this, unless someone's already got some work in progress here.

cryptojedi:
Douglas Stebila <notifications@github.com> wrote:
> I have some time today and could work on this, unless someone's
> already got some work in progress here.

I think I'm still entertained with setting up my server again and then
installing Windows in a VM.

thomwiggers:
I talked about this a little bit with Joost already and we at least identified the following points:

* It's probably helpful to have some shared python module to do some common things:
  * get a list of schemes
  * load a meta file for a scheme
  * Maybe other things, such as some common handling of arguments to e.g. set up logging

[...]

PR #25, opened 2019-02-12
Initial attempt at Python-driven checking framework

dstebila:
Demonstrates how we might address issue #10.

Main ideas:

- `scripts/PQClean.py` gets all the available schemes/implementations, and loads their metadata from the yaml file.
- `scripts/Tests.py` contains some helper logger routines, including colour-coded output (although not using the colorlog package as suggested in https://github.com/PQClean/PQClean/issues/10#issuecomment-462803162
- Individual tests appear in `scripts/check_*.py`, which follow the same pattern: a `run` function which takes a scheme name, and runs a check, using `Tests.report_success` or `Tests.report_error` to log the result and output
- `scripts/run_checks.py`, which runs all checks for all schemes (or can take command-line arguments to only run some checks on some schemes).  For successful checks, only a summary table is printed.  For failed checks, the full stdout/stderr capture is printed.

I've only implemented 3 checks so far as a demo:

- `check_license` which checks to make sure a LICENSE file is present for every implementation
- `check_compile_lib` which checks to make sure that each implementation compiles using its own Makefile without errors
- `check_functest` which compiles the functional tests for each scheme and checks that they run without errors
    - To make `check_functest` work, I added a `test/Makefile` which will build the requested functional test using against the lib*.a file that was compiled using the scheme's local Makefile

I tried to minimize the amount of repeated code that is present in individual `check_*.py` files, outsourcing it as much as possible to `run_checks.py`, `Tests.py`, and `PQClean.py`.

So far this runs on macOS and Linux.  Assuming Python is present on Windows, we might actually be able to get most of the checks running on Windows too, as long as we avoid platform-specific Python where possible, and then in ones like `check_compile_lib` we might be able to switch them over to using `nmake` rather than `make.

### DIFF
--- a/scripts/Tests.py
+++ b/scripts/Tests.py
@@ -1,0 +1,68 @@
+"""
+Utility functions for running tests and logging the results.
+"""
+
+import colorama
+import PQClean
+import subprocess
+
+colorama.init()
+
+globalreturnvalue = 0
+
+successes = []
+
+def report_success(check: str, scheme: PQClean.Scheme, implementation: PQClean.Implementation, capture: str = None):
+    """Report a successful test.  Note that successful test results are not immediately output, and instead are saved for output via print_successess."""
+    global successess
+    entry = {'check': check, 'scheme': scheme, 'implementation': implementation}
+    if entry not in successes: successes.append(entry)
+
+def print_successes():
+    """Print table listing all successful tests."""
+    global successes
+    print(colorama.Fore.GREEN + "=======================SUCCESSES=========================")
+    print("{:18}, {:18}, {:15}".format("check", "scheme", "implementation"))
+    print("---------------------------------------------------------")
+    for entry in successes:
+        print("{:18}, ".format(entry["check"]), end='')
+        print("{:18}, ".format(entry["scheme"].name if entry["scheme"] else ""), end='')
+        print("{:15}".format(entry["implementation"].name if entry["implementation"] else ""))
+    print("=========================================================")
+    print(colorama.Style.RESET_ALL)
+
+def report_failure(check: str, scheme: PQClean.Scheme, implementation: PQClean.Implementation, message: str, capture: str = None):
+    """Report a failed test.  Note that failed test results are immediately output."""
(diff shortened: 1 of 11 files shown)


scripts/Tests.py
```diff
@@ -0,0 +1,68 @@
+"""
```
thomwiggers: Nit: python modules should generally be lowercased.

scripts/run_checks.py
```diff
@@ -0,0 +1,48 @@
...
+import sys
+import PQClean
+import Tests
+
+# populate all checks from files named check_*.py, and import those modules
+all_check_names = []
+for check in glob.iglob('check_*.py'):
+    check = check[6:] # remove check_ prefix
+    check = check[:-3] # remove .py extension
+    exec("import check_" + check)
```
thomwiggers: `exec` is generally not safe. It's better to use `importlib` here:

```python
globals()[name] = importlib.import_module(name)
```

scripts/run_checks.py
```diff
@@ -0,0 +1,48 @@
...
+
+all_schemes = PQClean.Scheme.all_schemes()
+all_scheme_names = all_schemes.keys()
+
+def run(check_names: list, scheme_names: list):
+    """Run the requested checks for the requested schemes."""
+    global all_schemes
+    for scheme in scheme_names:
+        for check in check_names:
+            eval("check_" + check).run(all_schemes[scheme])
```
thomwiggers: Same for `eval`. You can instead use:

```python
globals()["check_"+check).run(all_schemes[scheme])
```

scripts/Tests.py
```diff
@@ -0,0 +1,68 @@
...
+Utility functions for running tests and logging the results.
```
henrydcase: why not to reuse ``unittest`` module. Something like was done here:
https://github.com/cloudflare/tls-tris/blob/master/_dev/interop_test_runner

thomwiggers:
Closed in favour of #27

PR #26, opened 2019-02-14
adds tests on arm7 and arm8

### LINKED ISSUE #18: CI runs for ARM
I think it could be interesting to CI runs for ARM. Goals being - ensure software can be easly ported to ARM CPUs and test against another compiler.

I can take care of it if people think it makes sense.

henrydcase:
Don't merge yet

Closes #18

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,24 @@ matrix:
         - docker run --rm --privileged multiarch/qemu-user-static:register --reset
         - docker run --rm -v `pwd`:`pwd` -w `pwd` "pqclean/debian-unstable-ppc" /bin/bash -c "uname -a &&
           make ${MAKETARGET}"
+    - name: "Run tests on qemu-arm32 (GCC)"
+      os: linux
+      services: docker
+      env:
+        - MAKETARGET="run-functest-all run-testvectors-all run-sanitizer-all run-symbol-namespace-all"
+      script:
+        - docker run --rm --privileged multiarch/qemu-user-static:register --reset
+        - docker run --rm -v `pwd`:`pwd` -w `pwd` "pqclean/debian-buster-arm" /bin/bash -c "uname -a &&
+          make ${MAKETARGET}"
+    - name: "Run tests on qemu-aarch64 (GCC)"
+      os: linux
+      services: docker
+      env:
+        - MAKETARGET="run-functest-all run-testvectors-all run-sanitizer-all run-symbol-namespace-all"
+      script:
+        - docker run --rm --privileged multiarch/qemu-user-static:register --reset
+        - docker run --rm -v `pwd`:`pwd` -w `pwd` "pqclean/debian-buster-aarch64" /bin/bash -c "uname -a &&
+          make ${MAKETARGET}"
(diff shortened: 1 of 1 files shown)


henrydcase:
`` pqclean/debian-buster-arm:latest not found``

thomwiggers:
Oops, the user travis pushes to hub as, did not have permissions. Should be fixed soon.

thomwiggers:
This tests both ARM and ARM8 (Aarch64)

.travis.yml
```diff
@@ -51,6 +51,24 @@ matrix:
...
         - docker run --rm -v `pwd`:`pwd` -w `pwd` "pqclean/debian-unstable-ppc" /bin/bash -c "uname -a &&
           make ${MAKETARGET}"
+    - name: "Run tests on qemu-arm32 (GCC)"
+      os: linux
+      services: docker
+      env:
+        - MAKETARGET="run-functest-all run-testvectors-all run-sanitizer-all run-symbol-namespace-all"
+      script:
+        - docker run --rm --privileged multiarch/qemu-user-static:register --reset
+        - docker run --rm -v `pwd`:`pwd` -w `pwd` "flowher/debian-buster-arm" /bin/bash -c "uname -a &&
```
thomwiggers: This should now be `pqclean/debian-buster-arm`

.travis.yml
```diff
@@ -51,6 +51,24 @@ matrix:
...
+          make ${MAKETARGET}"
+    - name: "Run tests on qemu-aarch64 (GCC)"
+      os: linux
+      services: docker
+      env:
+        - MAKETARGET="run-functest-all run-testvectors-all run-sanitizer-all run-symbol-namespace-all"
+      script:
+        - docker run --rm --privileged multiarch/qemu-user-static:register --reset
+        - docker run --rm -v `pwd`:`pwd` -w `pwd` "flowher/debian-buster-aarch64" /bin/bash -c "uname -a &&
```
thomwiggers: This should now be `pqclean/debian-buster-aarch64`

.travis.yml
henrydcase: yup, will change

thomwiggers:
Wait, what? Somehow travis hadn't run this...

henrydcase:
argh.. yes. I'm trying to debug it (it doesn't hang on my PC for some reason).
Anyway, I should have fix today for 32 and 64 today

BTW: I think  output binaries could be placed in architecture dependent locations (something like ``bin/arm64/stuff``.

henrydcase:
you know what - let's temporarily disable sanitizers in arm64 and 32. I'll do PR

dstebila:
> BTW: I think output binaries could be placed in architecture dependent locations (something like `bin/arm64/stuff`.

Fine, but we need to have a consistent way of determining the name for scripts.  In Python there's the [platform module](https://docs.python.org/3/library/platform.html); which of the methods in there are appropriate, if any?

henrydcase:
looking at sources for "platform" module I would say ``.machine()`` method. From sources it seems platform uses ``os.uname()``, which should give consistent naming on linux and mac.

PR #30, opened 2019-02-14
Implement symbol namespace check using new Python testing framework

dstebila:


test/test_symbol_namespace.py
```diff
@@ -0,0 +1,46 @@
...
+        *_, symtype, symbol = symbolstr.split()
+        if symtype in 'TR':
+            if not symbol.startswith(namespace) and not symbol.startswith('_' + namespace):
+                non_namespaced.append(symbol)
+
+    if non_namespaced:
+        print("Missing namespace literal {}".format(namespace))
+        for symbol in non_namespaced:
+            print("\ttype: {}, symbol: {}".format(symtype, symbol))
+        assert(False)
```
henrydcase: assert(False)?

test/test_symbol_namespace.py
```diff
@@ -0,0 +1,46 @@
...
+        implementation.path()
+    )
+    out = helpers.run_subprocess(
+        ['nm', '-g', 'lib{}_{}.a'.format(scheme_name, implementation_name)],
+        implementation.path() 
+    )
+
+    lines = out.strip().split("\n")
+    symbols = []
+    for line in lines:
```
henrydcase: pythonic way would be:
```
symbols = list(itertools.filterfalse(lambda x: not (' T ' in x or ' D ' in x or ' S ' in x) , lines))
```

test/test_symbol_namespace.py
thomwiggers: That makes sense, if there are any non-namespaced symbols this check fails.

test/test_symbol_namespace.py
thomwiggers: This could remain an iterator and be moved into line 36, but let's do that some other time.

PR #32, opened 2019-02-15
arm: temporarily disable sanitizers

henrydcase:
Seems docker has a problem with sanitizers on ARM. Goal is to fix the build and work on a sanitizer problem separately


henrydcase:
let see if it works

PR #40, opened 2019-02-18
Migrate test vector checks to pythonic testing framework

dstebila:


test/Makefile
```diff
@@ -9,20 +9,26 @@ SCHEME_DIR="../crypto_$(TYPE)/$(SCHEME)/$(IMPLEMENTATION)"
...
-	$(CC) $(CFLAGS) -DPQCLEAN_NAMESPACE=PQCLEAN_$(SCHEME_UPPERCASE) -I$(SCHEME_DIR) crypto_$(TYPE)/functest.c $(COMMON_FILES) -o $@ -L$(SCHEME_DIR) -l$(SCHEME)_$(IMPLEMENTATION)
+	$(CC) $(CFLAGS) -DPQCLEAN_NAMESPACE=PQCLEAN_$(SCHEME_UPPERCASE) -I$(SCHEME_DIR) crypto_$(TYPE)/testvectors.c $(COMMON_FILES) $(COMMON_DIR)/randombytes.c -o $@ -L$(SCHEME_DIR) -l$(SCHEME)_$(IMPLEMENTATION)
+
+$(DEST_DIR)/testvectors_$(SCHEME)_$(IMPLEMENTATION): rebuild-scheme crypto_$(TYPE)/testvectors.c $(COMMON_FILES) $(COMMON_DIR)/notrandombytes.c
+	mkdir -p $(DEST_DIR)
+	cd $(SCHEME_DIR) && make clean && make
+	$(CC) $(CFLAGS) -DPQCLEAN_NAMESPACE=PQCLEAN_$(SCHEME_UPPERCASE) -I$(SCHEME_DIR) crypto_$(TYPE)/testvectors.c $(COMMON_FILES) $(COMMON_DIR)/notrandombytes.c -o $@ -L$(SCHEME_DIR) -l$(SCHEME)_$(IMPLEMENTATION)
 
 clean:
 	$(RM) $(DEST_DIR)/functest_$(SCHEME)_$(IMPLEMENTATION)
```
thomwiggers: This should probably use `-` to suppress errors, otherwise if you run `make clean` twice it will fail (but still be in the desired state).

Additionally, `clean` should be a `.PHONY` target.

thomwiggers: The `clean` target still has no `-`

```Makefile
clean:
    - rm bla
```
Makes sure failing `rm` stop it from executing.

dstebila: It doesn't need it. $(RM) expands to `rm -f` which will not fail even if the file doesn't exist.

thomwiggers: Ah, TIL

thomwiggers:
Can be merged once tests pass (unfortunately github can't do that for me – I'm heading home)

Issue #15, opened 2019-02-05
Valgrind on PowerPC


thomwiggers:
@joostrijneveld did some experiments showing that it's possible to emulate (classic) PowerPC on Travis CI. However, it is not possible to run all of our tests. After some investigating, I have reached the following conclusions.

* QEMU has a "user" mode and a "system" mode of emulating architectures.
* `qemu-user-ppc` segfaults valgrind, probably because both qemu-user and valgrind intercept system calls and do other weird things.
* Everything works on the PowerBook G4 that we have, but we would need to integrate our own CI (and the typical containerisation solutions (Docker) aren't supported on 32-bit PowerPC).
* `qemu-system-ppc` (full virtual machine) works for valgrind, but is hard to automate.
* The MIPS-variant emulated by `qemu-user-mips` seems unsupposted by valgrind (and qemu-user would probably still not work).
* I'm having a hard time getting `qemu-user-ppc64` to work. Should not be confused with `ppc64le` or `ppc64el` which are little-endian! Likely to run into the same issues though.

The way forward seems to be to skip valgrind on PPC, like on OS X, though it would have been nice to perform the memory validations etc on the different architecture to make sure there are no endianness-assuming things that break memory somehow. Perhaps we can come back to it later – for now, the scripts we'll set up in #10 should skip qemu-ppc.

TL;DR: Emulating PPC in userspace breaks valgrind.

Possible solutions are:
* Use full system virtualisation with qemu-system. Need to worry about networking and going to be annoying to set up on Travis.
* Run CI runners on actual hardware.

For now: just skip Valgrind on ppc. Fix maybe someday.

dstebila:
For now, just skip Valgrind on ppc. For "second-tier" systems I think our initial goal would just be that the code builds successfully.

gkerde:
I've got a PowerPC Mac G5. Pretty fancy in it's day, but it still runs fine if you guys need some real hardware to test on

thomwiggers:
The issue is not so much the lack of hardware. The problem is more that there is no convenient build service that allows us to dynamically provision and tear down the build environment, much like how Travis does this. There are of course no service providers who do this. We tried (and do) emulating on Travis, as described above, but there are issues with Valgrind.

[...]

PR #44, opened 2019-02-19
Enable address and UB sanitizers

### LINKED ISSUE #33: ARM32/64: Enable sanitizers
There are two problems related to executing sanitizers on ARM builds
* ARM32 - hangs when executing Dillitium with sanitizers enabled (reason unknown, investigation ongoing)
* ARM64 - execution fails: before starting execution sanitizers spawns some thread for tracing execution. this is done under 'ptrace' which requires privileged permissions in docker. For some reason starting docker with ``docker  --privileged --cap-add SYS_PTRACE`` doesn't help. (ongoing)

henrydcase:
This enables address sanitizer and undefined behaviour sanitizer,
but disables leak detection sanitizer.
It seema there is an issue in qemu that we are using, which prevents
spawning leaksanitizer's tracer thread. But as leak detection is
already done by valgrind on x86 build, this functionality is not
absolute need at the moment.

Closes #33

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,21 +57,21 @@ matrix:
       os: linux
       services: docker
       env:
-        - MAKETARGET="run-functest-all"
+        - MAKETARGET="run-functest-all run-sanitizer-all"
       script:
         - docker run --rm --privileged multiarch/qemu-user-static:register --reset
         - docker run --rm -v `pwd`:`pwd` -w `pwd` "pqclean/debian-buster-arm" /bin/bash -c "uname -a &&
-          make ${MAKETARGET} &&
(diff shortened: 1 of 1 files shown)


thomwiggers:
Need to be careful to consider setting this for the qemu architectures when we merge the sanitizer builds into the pythonized testing framework.

henrydcase:
yes. I'll take a look at how to get it fixed in qemu

thomwiggers:
Valgrind has similar problems under qemu-user, I think it's maybe unavoidable due to how it works: doing weird things with system calls.

Qemu-system has no issues with valgrind.

thomwiggers:
(semi-related to #15)

PR #48, opened 2019-02-26
Add test to check that every .c / .h file triggers a library rebuild

### LINKED ISSUE #45: Test Makefile dependencies
Should make sure that the Makefiles really have dependencies on all .c and .h files
and possibly the Makefile itself.

dstebila:
Fixes #45

### DIFF
--- a/test/helpers.py
+++ b/test/helpers.py
@@ -1,7 +1,7 @@
 import subprocess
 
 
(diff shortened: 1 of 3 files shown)


thomwiggers:
Seems nosetests has a parallelism disabled by default, so all is well.

PR #58, opened 2019-02-27
Greatly simplify container builds

thomwiggers:
By just building everything in a container, the build script is much simpler.

This does expect us to merge the checks that are currently still in the makefile to the python scripts soon, because they are not run by this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: c
 
 dist: xenial
+language: minimal
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,75 +4,6 @@ dist: xenial
 
 matrix:
   include:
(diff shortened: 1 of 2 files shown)


thomwiggers:
Build will fail probably, as not all containers are available yet.

thomwiggers:
@joostrijneveld I tried out a competing CI service, [CircleCI](https://circleci.com) using the fork on my own account, and it does seem that builds there are significantly quicker. It's ~2x as quick per build and it runs four builds in parallel, which speeds up things a lot.

See https://github.com/thomwiggers/PQClean/blob/containerized-build-circleci/.circleci/config.yml
 and https://circleci.com/gh/thomwiggers/PQClean. The interface of CircleCI is a bit more confusing though...

PR #59, opened 2019-02-27
Faster check of makefile dependencies

thomwiggers:
We do not actually need to compile the files. Instead, we can use `make -q` to report if the scheme needs rebuilding.

### DIFF
--- a/test/test_makefile_dependencies.py
+++ b/test/test_makefile_dependencies.py
@@ -3,11 +3,9 @@ Checks that every .c and .h file in an implementation is present as a
 dependency of that scheme's Makefile.
 """
 
-import hashlib
 import os
 import pqclean
 import helpers
-import subprocess
 import glob
 import datetime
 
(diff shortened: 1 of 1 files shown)


test/test_makefile_dependencies.py
```diff
@@ -41,22 +41,24 @@ def check_makefile_dependencies(scheme_name, implementation_name, file):
...
+                           expected_returncode=0)
 
     # touch the candidate .c / .h file
     helpers.run_subprocess(['touch', '-t', now.strftime(formatstring), file])
 
-    # rebuild
-    helpers.run_subprocess(['make'], implementation.path())
+    # check if it needs to be rebuilt using make -q
+    helpers.run_subprocess(['make', '-q'], implementation.path(),
+                           expected_returncode=1)
```
dstebila: Oh, this is much cleverer!

PR #60, opened 2019-02-27
Add formatting tests to the python-based testing framework.

thomwiggers:
See #10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ matrix:
         - pip3 install -r requirements.txt
       script:
         - "cd test && python3 -m nose --rednose --verbose"
+      addons:
+        homebrew:
+          packages:
+            - astyle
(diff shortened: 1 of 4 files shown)


joostrijneveld:
This currently re-implements `make apply-format` in Python, and removes the `make format` target. I think it would still be convenient for scheme integrators / maintainers to be able to easily run the required formatting command.

Note that this also appears to pull in the changes from #59 (wrt the Makefile dependencies), but those are already in master as well.

thomwiggers:
I've rebased on Master. It doesn't do `apply-format`, but rather `check-format`.

I agree that we should probably include some way to run the formatter, but indeed documentation may be the way forward there.

thomwiggers:
I'm rebasing this on the #58 container-based build infrastructure, as that should install a more recent version of `astyle`.

PR #61, opened 2019-02-27
Set windows warnings to W4

thomwiggers:


thomwiggers:
The `/W4` flag seems to have a large overlap with `-Wconversion`, but not completely. Somehow `/W4` seems quite happy with the Kyber768 build on Windows, while `-Wconversion` is not.

thomwiggers:
Spliced the comment about `-Wconversion` into #63

PR #64, opened 2019-02-28
Valgrind in python

thomwiggers:
Moves Valgrind into the python-based testing framework.

Builds on #53 because I use the new `helper.make` function.


thomwiggers:
The CircleCI error is expected as #65 isn't merged yet.

thomwiggers:
Oops, there is no Valgrind in the Docker containers.

PR #65, opened 2019-02-28
Move builds to CircleCI build service

thomwiggers:
It's quite a lot faster than Travis, and shortens 30min pipelines to 5 minutes.


dstebila:
Wow, that's really speedy!

PR #70, opened 2019-03-04
Windows tests

### LINKED ISSUE #49: Run python tests on Windows/Appveyor
Follow-up on #9, we should also run the python tests on Windows. Python is already pre-installed on Appveyor, see https://www.appveyor.com/docs/windows-images-software/#python.

thomwiggers:
Runs the python testing framework on Windows.

PR is based on #68.

Closes #49


thomwiggers:
Appveyor fails because the `Makefile.Microsoft_nmake` files don't specify header dependencies...

thomwiggers:
I've managed to fix the slowness of the Windows-equivalent of `touch`, by just avoiding it: we now set the modification time via `os.utime`.

test/test_format.py
```diff
@@ -12,8 +12,10 @@ def test_formatting():
 
 
 def check_format(implementation: pqclean.Implementation):
-    cfiles = glob(os.path.join(implementation.path(), '*.c'))
-    hfiles = glob(os.path.join(implementation.path(), '*.h'))
+    if shutil.which('astyle') is None:
+        raise unittest.SkipTest("AStyle is not installed")
```
dstebila: Do we want to skip if astyle is not found or report a failure?  If our CI systems suddenly don't have astyle, we would want to know these are failing, not being skipped.

test/test_functest.py
```diff
@@ -24,21 +24,26 @@ def test_functest_sanitizers():
...
-    helpers.make(TYPE=implementation.scheme.type,
+    helpers.make(#'functest',
+                 TYPE=implementation.scheme.type,
                  SCHEME=implementation.scheme.name,
                  IMPLEMENTATION=implementation.name,
                  working_dir=os.path.join('..', 'test'))
     helpers.run_subprocess(
-        ['./functest_{}_{}'.format(implementation.scheme.name,
-                                   implementation.name)],
+        [os.path.join('..', 'bin', 'functest_{}_{}{}'.format(
```
dstebila: Shouldn't need the ../bin added in the command, as that's been set as the working dir on line 38 below.

test/test_functest.py
```diff
@@ -51,8 +56,11 @@ def check_functest_sanitizers(implementation):
                  working_dir=os.path.join('..', 'test'),
                  env=env)
     helpers.run_subprocess(
-        ['./functest_{}_{}'.format(implementation.scheme.name,
-                                   implementation.name)],
+        [os.path.join('..', 'bin', 'functest_{}_{}{}'.format(
```
dstebila: Shouldn't need the ../bin added in the command, as that's been set as the working dir on line 64 below.

test/test_testvectors.py
```diff
@@ -16,15 +16,19 @@ def test_testvectors():
...
-    helpers.make(TYPE=implementation.scheme.type,
+    helpers.make('testvectors',
+                 TYPE=implementation.scheme.type,
                  SCHEME=implementation.scheme.name,
                  IMPLEMENTATION=implementation.name,
                  working_dir=os.path.join('..', 'test'))
     out = helpers.run_subprocess(
-        ['./testvectors_{}_{}'.format(implementation.scheme.name,
-                                      implementation.name)],
+        [os.path.join('..', 'bin', 'testvectors_{}_{}{}'.format(
```
dstebila: Shouldn't need the ../bin added in the command, as that's been set as the working dir on line 30 below.

test/test_functest.py
thomwiggers: That's a bug or difference in behaviour in Python on Windows and Linux, which is not documented, but without `../bin/` this does not work on Windows, *even* if you supply `cwd`.

dstebila: Sigh.

test/test_format.py
thomwiggers: I guess I could add some detection for the CI environment, I did not want to push annoying failures on people in places where it's hard to install these tools (windows)

thomwiggers: See https://github.com/PQClean/PQClean/pull/70/commits/b2c8a136607ddd7c696c185c47b7b016f8548a01

thomwiggers:
It turned out that my installing astyle did not work as intended; I'll fix that and rebase the history into a nicer shape when done.

thomwiggers:
I learnt some new expletives to call windows, but I got it to work in the end. Ready for merging.

Issue #73, opened 2019-03-06
Incremental hash API


joostrijneveld:
As a consequence of #69, we need an incremental hash API: if we do not have the `sm` buffer available, we cannot keep the message in place and prefix it with other values as input to a hash function.

This means maintaining a slightly more complex state. In particular, for SHAKE, we need to be careful to not allow mixing of absorbs and squeezes. Block-level incremental squeezing is significantly easier, but insufficient (as that would put the burden of byte-wise increments on the scheme implementer).

PR #74, opened 2019-03-07
Add incremental API for SHAKE and SHA3

joostrijneveld:
This takes a step towards fixing #73. That issue also requires incremental SHA256, but that is less urgent; the Keccak variants suffice for most schemes. Merging this should be sufficient to move #73 to beyond 'accepting submissions.'


test/Makefile
```diff
@@ -17,7 +17,7 @@ DEST_DIR=../bin
 # This -Wall was supported by the European Commission through the ERC Starting Grant 805031 (EPOQUE)
 CFLAGS=-Wall -Wextra -Wpedantic -Werror -std=c99 -I$(COMMON_DIR) $(EXTRAFLAGS)
 
-all: $(DEST_DIR)/functest_$(SCHEME)_$(IMPLEMENTATION) $(DEST_DIR)/testvectors_$(SCHEME)_$(IMPLEMENTATION)
+all: $(DEST_DIR)/functest_$(SCHEME)_$(IMPLEMENTATION) $(DEST_DIR)/testvectors_$(SCHEME)_$(IMPLEMENTATION) $(DEST_DIR)/test_fips202
```
thomwiggers: I don't think we need to add this here.

PR #75, opened 2019-03-11
Add incremental SHA2

### LINKED ISSUE #73: Incremental hash API

joostrijneveld:
Currently we only offer SHA384 and SHA512. This adds SHA2, as well as an incremental API for all SHA2 variants. Closes #73.

It's block-level rather than byte-level (which is the case for SHA3), but that's a bit more tricky. Users of an incremental SHA2 API will likely be aware of block-level hashing anyway.


common/sha2.h
```diff
@@ -1,7 +1,27 @@
 #ifndef SHA2_H
 #define SHA2_H
 
-int sha384(uint8_t *out, const uint8_t *in, size_t inlen);
-int sha512(uint8_t *out, const uint8_t *in, size_t inlen);
+#include <stddef.h>
+#include <stdint.h>
+
+void sha224_inc_init(uint8_t *state);
+void sha224_inc_blocks(uint8_t *state, const uint8_t *in, size_t inblocks);
```
thomwiggers: these blocks are 64 bytes? Is that documented somewhere?

joostrijneveld: Good point - I've added a few lines of documentation.

PR #76, opened 2019-03-19
Namespace the #define constants.

thomwiggers:
This adds the namespace to the ``#define`` constants.
I'm also printing the ``CRYPTO_ALGNAME`` constant, because that then will make sure that it is also namespaced.


thomwiggers:
Having some issues with ``API_H``, created https://stackoverflow.com/questions/55243068/check-if-a-variable-constant-is-defined

joostrijneveld:
Perhaps this is also a good moment to look at the `#include params.h`; is there anything reasonable we can do to check for those kind of defines (or, alternatively, any includes in `api.h` at all)? I don't immediately see any other option than manual checking..

PR #78, opened 2019-04-01
Add FrodoKEM-640-SHAKE reference implementation

dstebila:


joostrijneveld:
I'm not sure how strict we want to be on 'integers should be of fixed size'. Concretely, I see several occurrences of `int i` and some `unsigned char` byte arrays.

Perhaps this should go in a separate issue, but maybe we should reconsider automatic checks for those. I briefly looked into [pycparser](https://github.com/eliben/pycparser) and I'm positively surprised how easy it is to get a decent abstract syntax tree.

Issue #79, opened 2019-04-02
Automatically check for char / int


joostrijneveld:
As I mentioned in https://github.com/PQClean/PQClean/pull/78#issuecomment-479112814, it turns out to be quite tedious to manually find all non-fixed-size integer types. Perhaps we should consider using something like [pycparser](https://github.com/eliben/pycparser) to search the syntax tree. I played around a bit based on the code in #78, and it turns out to be fairly simple to produce AST fragments that find these issues, e.g.,

```
[...]
      For: 
        DeclList: 
          Decl: i, [], [], []
            TypeDecl: i, []
              IdentifierType: ['int']
```

for [this for-loop](https://github.com/PQClean/PQClean/pull/78/files#diff-522f5174363c09963e6f79b74f7a7c55R213), or e.g. [this function declaration](https://github.com/PQClean/PQClean/pull/78/files#diff-25d475e5cf696f6722d3639cbf5eb6b1R128):

```
  Decl: PQCLEAN_FRODOKEM640SHAKE_CLEAN_pack, [], [], []
    FuncDecl: 
      ParamList: 
        Decl: out, [], [], []
          PtrDecl: []
            TypeDecl: out, []
              IdentifierType: ['unsigned', 'char']
        Decl: outlen, [], [], []
          TypeDecl: outlen, []
            IdentifierType: ['size_t']
[...]
```

This means we would really draw a hard line and make CI fail for any of these types.

Is this something we want to implement (perhaps before explicitly opening up for submissions)?

(Tangentially related to the discussion in #77)

dstebila:
Will taking such a hard line prevent potential security flaws?  Or will it just annoy people trying to port their code to PQClean?

joostrijneveld:
Honestly, I think it'll mostly end up annoying people, but I'm already somewhat leaning towards that opinion on `unsigned char` vs `uint8_t` altogether. I find it hard to judge on an individual, manual code-review basis, though - in that sense having a hard failing CI test is a clear way of drawing the line.

Did you feel mostly annoyed when doing those changes for Frodo, or did it seem sensible? For SPHINCS+, I feel like I'm mostly not looking forward to the long namespace strings, and not so much the changing to fixed-width types.

dstebila:
For Frodo there were only a few places where I had to change it, and it wasn't that annoying.  I suppose it made the code more consistent, but it also didn't feel like it made the code quantifiably more secure, just that it might be easier for a human to read later because of the consistency.

[...]

PR #82, opened 2019-04-04
Prohibit using char without explicit sign modifier

joostrijneveld:
Related to #79; compilers are free to choose whether `char` is signed or unsigned.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ matrix:
         - pip3 install -r requirements.txt
         - export PATH="/usr/local/bin:$PATH"
         - ln -s /usr/local/Cellar/gcc/8.3.0/bin/gcc-8 /usr/local/bin/gcc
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -20,9 +20,11 @@ matrix:
         homebrew:
           packages:
             - astyle
-            - gcc@8
+            - gcc
```
joostrijneveld: Why do we not explicitly ask for version 8 anymore? Does this not break when homebrew exposes gcc 8.3.1 as `gcc`, as our symlink is pinned to 8.3.0? This implies that we may even want to pin to 8.3.0..

Is this whole mess not resolved by doing `brew link` instead of manually creating a symlink to the Cellar?

joostrijneveld: This was resolved after some offline discussion, in eb08730d277f61e0344c35c9e450fb86fb54503a. By linking to `/usr/local/bin/gcc-8` we are agnostic to minor versions, and by pinning to `gcc@8`, we prevent accidentally bumping a major version.

Issue #81, opened 2019-04-03
How to deal with code duplication?


dstebila:
Thinking forward to adding multiple parameterizations of algorithms (e.g. #80) or multiple implementations of the same algorithm / parameterization (e.g., clean C, AVX2, ...):

There will probably be a lot of duplication.  If I think about Frodo or Kyber or probably others, the clean implementations of different security levels are exactly the same, except for the params.h file which contains the compile-time macros defining various sizes.  

- Do we want to have duplicate .c files sprayed across the repository?  
- How will we ensure consistency between them?  

And if we start to do optimized implementations in other subdirectories, we probably have even more duplication, as in many cases it may be only one or two files that have been heavily optimized, the rest unchanged.

For example, in Frodo, there are 6 parameterizations {level 1, level 3, level 5} x {AES, SHAKE for matrix generation}.  All files save params.h and matrix.c are the same.

dstebila:
I suppose one solution to ensuring consistency would be to add some metadata somewhere saying which files are expected to be identical, and then add CI tests that check those files remain identical...

cryptojedi:
Douglas Stebila <notifications@github.com> wrote:

> I suppose one solution to ensuring consistency would be to add some
> metadata somewhere saying which files are expected to be identical,
> and then add CI tests that check those files remain identical...

How about using symlinks?

dstebila:
> How about using symlinks?

Possibly problematic.  

1. They don't work on Windows. https://github.com/git-for-windows/git/wiki/Symbolic-Links
2. When browsing the repository on Github, you see the links, not the contents of the files. https://github.com/PQClean/PQClean/tree/symlinks/crypto_kem/frodokem640aes/clean

cryptojedi:
Douglas Stebila <notifications@github.com> wrote:

> I've just realized it's even worse: all functions are namespaced
> according to the parameterization, so duplicate files are only
> identical modulo namespacing.

OK, that makes a lot of sense, so no symlinks.

dstebila:
I've been working on another security level of Frodo, and have tried adding a check for duplicate consistency across schemes as follows:

1. In the META.yml file, add a field called 'duplicate-consistency' listing what files should be consistent after namespace renaming. (https://github.com/PQClean/PQClean/blob/more-frodo/crypto_kem/frodokem976shake/META.yml#L23)
2. Then, a new test checks all stated consistency requirements. 
[...]

PR #83, opened 2019-04-04
Add FrodoKEM-976-SHAKE

dstebila:
Adds a "duplicate consistency" check to address #81


dstebila:
One CircleCI build was marked as failed, but that appears to be a quirk (it failed on a clang-tidy test for kyber768, completely unrelated to the changes here).  I re-ran the build and it passed, but it doesn't update the build status from red to green.

dstebila:
Ready for review. Don't need to code review most of the C files as they are duplicates of FrodoKEM-640-SHAKE.

PR #86, opened 2019-04-05
Use 5 iterations for functional tests and test vectors

### LINKED ISSUE #85: Reduce number of iterations
Reduce number of iterations in functest and testvectors to say 5.

dstebila:
Fixes #85.

## Manually checked properties

* [ ] `#ifdef`s only for header encapsulation
* [ ] `api.h` does not include other files
* Optional:

### DIFF
--- a/crypto_kem/frodokem640shake/META.yml
+++ b/crypto_kem/frodokem640shake/META.yml
@@ -3,7 +3,7 @@ type: kem
 claimed-nist-level: 1
 length-public-key: 9616
 length-ciphertext: 9720
-testvectors-sha256: 521ff891de20efe74e6584d09612dae989427ac76261a41630c4e4d6a4fc78a4
+testvectors-sha256: 8f922de02d41005fcc3c4164b2ab74c4c7b588ed69e34e22607d1ae4ab13d2c5
 principal-submitter: Douglas Stebila, University of Waterloo
 auxiliary-submitters:
   - Erdem Alkim
--- a/crypto_kem/kyber768/META.yml
+++ b/crypto_kem/kyber768/META.yml
@@ -3,7 +3,7 @@ type: kem
 claimed-nist-level: 3
 length-public-key: 1088
 length-ciphertext: 1152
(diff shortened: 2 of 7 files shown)


dstebila:
There's something weird in `test/crypto_sign/testvectors.c` which I wasn't sure how to fix.  The `NTESTS` macro isn't being used to control the number of tests, instead there's `MAXMLEN`.  @mkannwischer you wrote this originally, can you help?

joostrijneveld:
I just ran into the same thing w.r.t. detached signatures. I'll give it a shot.

dstebila:
I'm still confused -- what does that code actually do?

joostrijneveld:
I believe it generates several signatures, for increasing message lengths up to `MAXMLEN`, increasing the length exponentially.

dstebila:
> I believe it generates several signatures, for increasing message lengths up to `MAXMLEN`, increasing the length exponentially.

Makes sense, thanks!

joostrijneveld:
Hmm, this is probably not the right place for `gen_testvector_hash.py`; now nose picks it up as a test. That should probably live somewhere else, but that requires some edits w.r.t. relative imports. I'll look at that tonight/tomorrow.

thomwiggers:
> now nose picks it up as a test

I think there's some logic to that, if you change the names or make the file executable it should be ignored.

https://nose.readthedocs.io/en/latest/finding_tests.html

dstebila:
Do we really need gen_testvector_hash.py?  I just did 

    bin/testvectors_frodo640kemshake_clean | sha256sum

and copied the output.

cryptojedi:
> Do we really need gen_testvector_hash.py?  I just did 
> 
>     bin/testvectors_frodo640kemshake_clean | sha256sum
> 
> and copied the output.

I agree, that's probably the easiest thing to do.

joostrijneveld:
Right, I clearly didn't think this through and went autopilot based on `test/test_testvectors.py`; I removed the script again.

PR #87, opened 2019-04-05
Add remaining length values to META.yml

### LINKED ISSUE #84: Add lengths to META.yml
META.yml files currently include length-public-key and length-ciphertext.  They should also include length-secret-key and length-shared-secret (for KEMs).  Similarly for signatures.

dstebila:
Fixes #84

## Manually checked properties

* [ ] `#ifdef`s only for header encapsulation
* [ ] `api.h` does not include other files
* Optional:


thomwiggers:
Ah shit, I have just done this exactly and yolo-pushed it to master. Sorry about that!

dstebila:
No worries!

Issue #69, opened 2019-03-04
Implement and test detached signature API


thomwiggers:
In #55 we discovered that Dilithium does not implement `crypto_sign_signature` and `crypto_sign_verify`, but only `crypto_sign` and `crypto_sign_open`.

joostrijneveld:
Let's make this issue a bit broader, as we also need to implement functional tests for these API functions.

Can we even do meaningful test vector comparison, given that we currently do this by comparing to (a hash of) `sm` bytestreams? I don't immediately see a way to test for equivalence between `crypto_sign` and `crypto_sign_signature` without manually dissecting `sm`.

cryptojedi:
Joost Rijneveld <notifications@github.com> wrote:
> Let's make this issue a bit broader, as we also need to implement
> functional tests for these API functions.
> 
> Can we even do meaningful test vector comparison, given that we
> currently do this by comparing to (a hash of) `sm` bytestreams? I
> don't immediately see a way to test for equivalence between
> `crypto_sign` and `crypto_sign_signature` without manually dissecting
> `sm`.

Do we have to have equivalence there? It seems rather clear to me that
you'd use either one API or the other and not a mixture of both. As long
as each approach works, it's fine, right?

dstebila:
There's no guarantee that `crypto_sign_signature`'s output is even a substring of `crypto_sign`'s output, there could be some weird transformation or padding applied.

I think at this point all we can do is demand test vectors for both, and maybe add a unit test that checks the compile library files to make sure the relevant functions are in fact exported (which would have caught the fact that Dilithium was missing these).

thomwiggers:
Adding a separate check for the symbols is not necessary, these functions should just be included in `crypto_sign/functest.c`, where missing symbols would just turn into a linking error.

PR #90, opened 2019-04-05
SPHINCS-SHAKE256-128f-simple

joostrijneveld:
~This depends on #86.~ The main goal here was to create a test case for #69; the other 35 parameter sets of SPHINCS+ will follow once I've written a script to automatically do the namespacing..

~After #86 makes it in, it's probably best to rebase this on master, since it currently does not include the length required by https://github.com/PQClean/PQClean/commit/d39d7c66dbab4fa3497a2b7b9b65a011f308b121~ Done!

* [x] `#ifdef`s only for header encapsulation
* [x] `api.h` does not include other files
* Optional:


joostrijneveld:
This fails because of variable-length arrays on the stack. See e.g. 

https://github.com/PQClean/PQClean/blob/c0eeaf033a443a456399f2b3cfba84add3e346ab/crypto_sign/sphincs-shake256-128f-simple/clean/utils.c#L109-L110

and 

https://github.com/PQClean/PQClean/blob/c0eeaf033a443a456399f2b3cfba84add3e346ab/crypto_sign/sphincs-shake256-128f-simple/clean/thash_shake256_simple.c#L16

Both `treeheight` and `inblocks` are fixed at compile-time, and are function parameters simply to prevent function duplication; both the `thash` and `treehash` functions are called for various dimensions. I don't consider this an error, as it doesn't lead to potential memory leaks or stack overruns. Is there a way of dealing with this such that Visual Studio agrees? What's the C-before-VLA's pattern to do this?

thomwiggers:
> This does not pass CI on Windows because of variable-length arrays on the stack. [...]

I've found that VLAs are a [C99 feature that's even been made optional in C11][1]. Microsoft [does not support C99, only the features needed to implement C++][2].


[1]: https://en.wikipedia.org/wiki/Variable-length_array 
[2]: https://github.com/MicrosoftDocs/cpp-docs/issues/407#issuecomment-417922008

joostrijneveld:
I'm not sure how to interpret the '_`#ifdef` only for header encapsulation_' rule. Does that mean snippits like the fragment from `params.h`, below, is not allowed? What are we trying to prevent?

https://github.com/PQClean/PQClean/blob/e7dacfb723e690189fb82653a273cc36d7dd6f0c/crypto_sign/sphincs-shake256-128f-simple/clean/params.h#L33-L42

joostrijneveld:
After some offline discussion with @mkannwischer and @cryptojedi, we concluded:

- we can easily flatten `params.h` for the individual parameter sets, removing all preprocessor code
- we get rid of variable length arrays. `treehash` is called with 2 different lengths (FORS and subtrees), `thash` with 4 (1, 2, the number of FORS trees and the number of WOTS chains). We'll use small wrappers for each of these calls that allocate fixed-length buffers and pass those to the generic functions.

I'll finish this later today.

PR #95, opened 2019-04-09
Aes

cryptojedi:
Added Code for AES128, AES192, and AES256 in ECB and CTR mode.

* [x] `#if`/`#ifdef`s only for header encapsulation
* [x] `api.h` does not include other files
* Optional:


joostrijneveld:
What would be a good resource for test vectors to compare these functions to? It would be nice to include a functional test in `test/common/`.

mkannwischer:
Shouldn't the test return 1 on errors?

thomwiggers:
The AES tests should also be added to the `tests/Makefile` to build them, and we should probably also integrate them in the python testing framework somehow.

PR #100, opened 2019-04-09
Add test to cross-check OBJECTS between Makefiles

### LINKED ISSUE #99: We might want to add a test that Makefile and Makefile.Microsoft_nmake want the same object files

joostrijneveld:
Fixes #99

### DIFF
--- a/test/test_crosscheck_makefile_objects.py
+++ b/test/test_crosscheck_makefile_objects.py
@@ -1,0 +1,47 @@
+"""
+Catch-early check that sees if OBJECTS matches across Makefile and NMakefile
+"""
+
+import os
+import pqclean
+
+
+def test_makefile():
+    for scheme in pqclean.Scheme.all_schemes():
+        for implementation in scheme.implementations:
+            yield check_makefile, implementation
+
+
+def check_makefile(implementation):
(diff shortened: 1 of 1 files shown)


thomwiggers:
This does mean that you're hardcoding that people need to define exactly OBJECTS as a variable.

joostrijneveld:
I don't have a problem with that - in fact, I wouldn't mind enforcing a much stronger template on the Makefile, i.e., essentially only let implementations differ in `HEADERS` and `OBJECTS`.

Briefly talked to @cryptojedi about this yesterday; the main argument against including the common build targets is to keep the makefiles self-contained -- I guess we've all seen a project or two where you find yourself to tracing Makefile-includes for half an hour.

dstebila:
Do we really need this?  There's already a check that every .c / .h file is used by Makefile, and both Makefile and Makefile.Microsoft_nmake are exercised by the CI system.  I don't see any security risk arising from these two Makefiles being slightly out-of-sync.  So to me this could be upping the annoyance factor for contributors.

joostrijneveld:
The motivation that led up to this was a typo that this test would've caught when developing on Linux and only testing on Windows on CI.. but yeah, in hindsight I agree that it's clutter. These errors will be caught by CI, and that's fine.

PR #101, opened 2019-04-10
Add FrodoKEM AES parameterizations

### LINKED ISSUE #80: Add remaining FrodoKEM parameterizations

dstebila:
Fixes #80.  Depends on #95.

Enhances duplicate\_consistency checks to allow checking against multiple sources (in Frodo's case, we want to check the consistency of parts of FrodoKEM-640-AES with FrodoKEM-976-AES and FrodoKEM-640-SHAKE.

* Optional:


test/test_valgrind.py
```diff
@@ -21,6 +21,8 @@ def check_valgrind(implementation: pqclean.Implementation):
     if (platform.machine() not in ('i386', 'x86_64') or
             platform.system() != 'Linux'):
         raise unittest.SkipTest()
+    if os.environ.get('CIRCLECI') == 'true' and os.environ.get('ARCH') == 'i386' and (implementation.scheme.name == 'frodokem1344aes' or implementation.scheme.name == 'frodokem1344shake'):
+        raise unittest.SkipTest('Skipping too-slow valgrind test')
```
thomwiggers: Not sure if we want this all the time. For SPHINX it will be even worse - if valgrind proves too slow, we might want to just remove it from all CI tests.

Also, only disable it on CircleCI? What about Appveyor/Travis?

test/test_duplicate_consistency.py
```diff
@@ -9,30 +9,31 @@
 import yaml
 
 def test_duplicate_consistency():
+    helpers.skip_windows()
```
thomwiggers: This function is a decorator.

test/test_valgrind.py
dstebila: Yes we need to decide which tests we're running on which CI environments. But let's save that for an issue on its own.

test/test_duplicate_consistency.py
dstebila: I don't understand what that means.

thomwiggers: Ah, it means you can use it as the following:

```python
@helpers.skip_windows()
def test_common():
```

thomwiggers: I don't think it will skip windows if it's not used as a decorator.

test/test_valgrind.py
thomwiggers: But don't you intend to also not run the valgrind test on Travis and Appveyor? Shouldn't you do `'CI' in os.environ`?

dstebila: It wasn't running too slow on Travis and Appveyor -- it's only the qemu i386 CircleCI builds that it was running too slow on.

PR #104, opened 2019-04-11
test_format wasn't actually checking output of astyle

dstebila:
It was only checking the return code, but astyle was always returning 0. 
Changed to parse the output and look for the string "Formatted"

* Optional:


dstebila:
Not sure why one of the AppVeyor's didn't execute (maybe because I first marked it as a "draft pull request"?)  Other AppVeyor passed so it seems mergeable to me.

PR #105, opened 2019-04-11
Add ntruhps2048509

dstebila:
Based on the reference implementation in the ZIP file posted on the NIST Round 2 site.

Only logic change so far has been to turn compile-time #if's into runtime if's, please see commit 1f50b12 specifically.

Also, the original code had some #ifdef's deciding between NTRU_HRSS and NTRU_HPS.  I have stripped out those #ifdefs since this code is only HPS, and we'll have to add it back in when we replicate this directory for the HRSS scheme.

* Optional:


dstebila:
@joostrijneveld or @cryptojedi, can you help resolve some algorithmic things that I am not sure about?

1. AppVeyor flags several conversion problems, some of which are rather weird and I don't know how to resolve safely, for example, `cmov` has an `unsigned char` argument `b` and then has a line `b = -b` which the compiler rightfully complains about.  https://github.com/PQClean/PQClean/pull/105/files#diff-7723f64f7e1fe15b52e7016170de9d53R22
2. In the original crypto\_sort.c source file, line 38 (see attached screenshot) is indented as if it's in the for loop, but it is a for loop without braces, so the compiler does not treat it as in the for loop. The linter complained about this; I added braces such that it preserves the actual algorithm (i.e., that line is outside the for loop https://github.com/PQClean/PQClean/pull/105/files#diff-3e5617df950f2b6652d50e304c2a16caR45), but this is worth checking.

Screen Shot 2019-04-10 at 20 59 59

joostrijneveld:
> 1. AppVeyor flags several conversion problems, some of which are rather weird and I don't know how to resolve safely, for example, `cmov` has an `unsigned char` argument `b` and then has a line `b = -b` which the compiler rightfully complains about.

Hmm, the cmov thing comes from Kyber, where, strangely enough, [AppVeyor does not complain](https://ci.appveyor.com/project/PQClean/pqclean/builds/23717634/job/6wbyh0y131e02nqd).. @cryptojedi fixed this [in the Kyberv2 branch for the `verify` function](https://github.com/PQClean/PQClean/blob/kyberv2/crypto_kem/kyber512/clean/verify.c#L24) by explicitly doing `r = ~r +1` rather than `r = -r`, which should lead to the same results. Still, the [cmov there contains `b = -b`](https://github.com/PQClean/PQClean/blob/kyberv2/crypto_kem/kyber512/clean/verify.c#L45) (and passes CI). I suppose the same thing works for cmov?

> 2. In the original crypto_sort.c source file, line 38 is indented as if it's in the for loop, but it is a for loop without braces, so the compiler does not treat it as in the for loop. The linter complained about this; I added braces such that it preserves the actual algorithm (i.e., that line is outside the for loop), but this is worth checking.

It seems that this is the result of automatic tab-to-spaces conversion gone wrong. In `supercop-20190110`, where this file originates, that line contains a tab and two spaces whereas all surrounding lines are indented using spaces. There, setting the tab width to 8 spaces puts it in the right place (i.e. one level higher than line 37, outside the for loop).

dstebila:
This is ready for review.  The only build failure is the clang-tidy segfault that we're seeing elsewhere which I presume you are fixing in #106

crypto_kem/ntruhps2048509/clean/LICENSE
```diff
@@ -0,0 +1,15 @@
+Public Domain
```
dstebila: @cryptojedi or @joostrijneveld What's the right license for the NTRU code?

crypto_kem/ntruhps2048509/clean/LICENSE
```diff
@@ -0,0 +1,15 @@
...
+Authors: Joppe Bos,
```
dstebila: Wrong author set - this was copied from Kyber, needs to be fixed.

crypto_kem/ntruhps2048509/clean/Makefile
```diff
@@ -0,0 +1,19 @@
+# This Makefile can be used with GNU Make or BSD Make
+
+LIB=libntruhps2048509_clean.a
+HEADERS=api.h crypto_sort.h owcpa.h params.h poly.h sample.h verify.h
+OBJECTS=crypto_sort.o kem.o owcpa.o pack3.o packq.o poly.o sample.o verify.o
+
+CFLAGS=-Wall -Wextra -Wpedantic -Werror -std=c99 -I../../../common $(EXTRAFLAGS)
```
mkannwischer:
```suggestion
CFLAGS=-Wall -Wextra -Wpedantic -Werror -Wmissing-prototypes -std=c99 -I../../../common $(EXTRAFLAGS)
```

PR #107, opened 2019-04-11
Fix a few nits

dstebila:
I was getting a clang-tidy error in SPHINCS.  Also adding test skipping logic to test\_common.

* Optional:
  * [x] Integers used for indexing are of size `size_t`


joostrijneveld:
Thanks! Not sure why `clang-tidy` is not always picking up on these kind of problems; it didn't complain in the original pull request.

Issue #110, opened 2019-04-12
Should we also check against NIST submission KAT values?


dstebila:
Because we are using our own format for test vectors, and thereafter a hash of those, our testvector checks end up ensuring consistency of our implementations with themselves over time.  But we don't have any check against the "canonical" implementation fro the NIST submissions, so it could be that our first version (which we used to generate the hash of the test vectors) is already distinct/incompatible/wrong.  We should have some check against the "ground truth" established by the code in the NIST submissions.

joostrijneveld:
I think that's very useful. I see that you replaced the AES implementation that they depended on OpenSSL for with a call to our own; with this dependency resolved, I don't really see any reason for doing our own test vector generation anymore.

dstebila:
More rounds is probably good to have, but I also like CI tests that finish quickly. :)  Yes I suppose we could have multiple hashes.  On the one hand yes we could get rid of our own test vector generation, but I wasn't necessarily implying this, I just wanted to check that I hadn't screwed up any implementations I'd ported over.

dstebila:
Fixed by #111

PR #109, opened 2019-04-12
Add -Wmissing-prototypes

mkannwischer:
Fixes #108. 

We are currently not including `-Wmissing-prototypes`.

When turning it on, I found that:

- [address.c](https://github.com/PQClean/PQClean/blob/master/crypto_sign/sphincs-shake256-128f-simple/clean/address.c) of SPHINCS doesn't include address.h - which it should
- The functions `br_range_enc32le` and `br_aes_ct64_skey_expand` in [aes.c](https://github.com/PQClean/PQClean/blob/master/common/aes.c) should be static, but are not. 
-  Frodo's [noise.c](https://github.com/PQClean/PQClean/blob/master/crypto_kem/frodokem640shake/clean/noise.c) should include [common.h](https://github.com/PQClean/PQClean/blob/master/crypto_kem/frodokem640shake/clean/common.h)
- Frodo's [utils.c](https://github.com/PQClean/PQClean/blob/master/crypto_kem/frodokem640shake/clean/util.c) should include [common.h](https://github.com/PQClean/PQClean/blob/master/crypto_kem/frodokem640shake/clean/common.h)
- Kyber's [verify.c](https://github.com/PQClean/PQClean/blob/master/crypto_kem/kyber768/clean/verify.c) should include [verify.h](https://github.com/PQClean/PQClean/blob/master/crypto_kem/kyber768/clean/verify.h)
- Some functions in Frodo differ from their prototypes in the sense that the implementation has const non-pointer arguments whereas the prototype does not. gcc+clang are fine with this, but on Windows it breaks. Since those consts don't change anything, I've removed them. 

I've fixed all of those issues.

Issue #108, opened 2019-04-12
-Wmissing-prototypes


mkannwischer:
We are currently not including `-Wmissing-prototypes`. 

When turning it on, I found that: 
- [address.c](https://github.com/PQClean/PQClean/blob/master/crypto_sign/sphincs-shake256-128f-simple/clean/address.c) of SPHINCS doesn't include address.h - which it should
- The functions `br_range_enc32le` and `br_aes_ct64_skey_expand` in [aes.c](https://github.com/PQClean/PQClean/blob/master/common/aes.c) should be static, but are not. 
-  Frodo's [noise.c](https://github.com/PQClean/PQClean/blob/master/crypto_kem/frodokem640shake/clean/noise.c) should include [common.h](https://github.com/PQClean/PQClean/blob/master/crypto_kem/frodokem640shake/clean/common.h)
- Frodo's [utils.c](https://github.com/PQClean/PQClean/blob/master/crypto_kem/frodokem640shake/clean/util.c) should include [common.h](https://github.com/PQClean/PQClean/blob/master/crypto_kem/frodokem640shake/clean/common.h)
- Kyber's [verify.c](https://github.com/PQClean/PQClean/blob/master/crypto_kem/kyber768/clean/verify.c) should include [verify.h](https://github.com/PQClean/PQClean/blob/master/crypto_kem/kyber768/clean/verify.h)
- Frodo's prototypes of `LE_TO_UINT16` and `UINT16_TO_LE` are not matching the implementation.

Should we turn it on? I think it is useful.

mkannwischer:
Ok, this triggered some more problems:
The builds on Windows don't like that some prototypes differ in `const`-ness from the implementation.
If I add the `const`'s, clang-tidy complains about having top level `const`'s in declarations: https://clang.llvm.org/extra/clang-tidy/checks/readability-avoid-const-params-in-decls.html

Do we want to eliminate the top level consts entirely?

cryptojedi:
"Matthias J. Kannwischer" <notifications@github.com> wrote:
> Ok, this triggered some more problems:
> The builds on Windows don't like that some prototypes differ in `const`-ness from the implementation.
> If I add the `const`'s, clang-tidy complains about having top level `const`'s in declarations: https://clang.llvm.org/extra/clang-tidy/checks/readability-avoid-const-params-in-decls.html
> 
> Do we want to eliminate the top level consts entirely? 

Do I understand correctly that those are non-pointer arguments. If yes,
[...]

Issue #93, opened 2019-04-08
Secret key should not be part of test vectors


joostrijneveld:
This came up while discussing #92; we allow people to vary secret key lengths across implementations, but our test vectors are even stricter, requiring them to be byte-wise identical. We should probably just remove the secret key from the test vector output.

cryptojedi:
This is related to the discussion we had about test vectors (and to including secret-key size in the YAML file). Essentially it boils down to the question of when two schemes are different. We could decide that as soon as schemes produce different testvectors, including secret keys, they are different schemes. Then also the secret-key size can go back into the YAML file. Maybe that's the most consistent approach.

dstebila:
If we have a clean and an optimized implementation of the same scheme, are they obligated to produce the same test vectors? Use the same secret key format?

cryptojedi:
Douglas Stebila <notifications@github.com> wrote:

> If we have a clean and an optimized implementation of the same scheme,
> are they obligated to produce the same test vectors? Use the same
> secret key format?

That's sort of the question. For the moment, we can decide that the
answer is "yes" and then all kind of issues go away.

dstebila:
Focusing on KEMs, I think there are three levels of interoperability we could use to define "being the same scheme", in increasingly levels of stringency:

- public-message-interoperable: two implementations that process each others' ciphertexts and public keys arrive at the same shared secret
- secret-key-interoperable: public-message-interoperable, plus the fact that the decapsulation code from one implementation can use the secret key output by keypair generation from another implementation
- randomness-equivalent: two implementations produce exactly the same output when run with the same randomness

RSA in SSH or TLS libraries has "public message interoperability", but may store secret keys in different formats.

AES has secret-key-interoperability, albeit not entirely, as different implementations may represent an expanded key schedule in different ways.

I'd prefer to demand only public-message-interoperability, because I think implementers should be allowed to choose between trade-offs in secret key representation.

dstebila:
I think we've come to the conclusion that secret keys will remain part of the test vectors especially since they're part of the NIST KATs.  Any objections to closing this issue?

thomwiggers:
Nope, let's revisit this if the issue comes up.

PR #111, opened 2019-04-14
Add NIST KAT check for KEMs

dstebila:
This addresses #110. It adds a copy of the NIST KAT generation program, then checks that the outputs of that KAT program match a hash value.  The hash value should be generated directly from the KAT files in the original NIST submission.  

Note that it is currently only implemented for KEMs.  If people are happy with this, then we can add it for signatures.  

Note as well that the KAT values for our kyber768 implementation don't match the ones in the Round 2 submission.  If memory serves, this is because kyber768 was modified for Round after the version that was added here.  I added a line to temporarily skip this test for kyber768.

* Optional:


test/crypto_kem/nistkat.c
```diff
@@ -0,0 +1,95 @@
...
+
+    nist_kat_init(seed, NULL, 256);
+
+    rc = crypto_kem_keypair(public_key, secret_key);
+    if (rc != 0) {
+        fprintf(stderr, "[kat_kem] %s ERROR: crypto_kem_keypair failed!\n", CRYPTO_ALGNAME);
+        return -1;
+    }
+    fprintBstr(fh, "pk = ", public_key, CRYPTO_PUBLICKEYBYTES);
+    fprintBstr(fh, "sk = ", secret_key, CRYPTO_SECRETKEYBYTES);
```
thomwiggers: I see that the NIST KAT files still contain the secret keys, thus not solving the problems discussed in #93. Does the (original) KAT format address that problem at all? Or do we still need to think about a new format that specifies equivalence up to secret keys (whenever that issue comes up)

common/nistkatrng.c
```diff
@@ -0,0 +1,103 @@
...
+//    key - 256-bit AES key
+//    ctr - a 128-bit plaintext value
+//    buffer - a 128-bit ciphertext value
+static void AES256_ECB(uint8_t *key, uint8_t *ctr, uint8_t *buffer) {
+    aes256ctx ctx;
+    aes256_keyexp(&ctx, key);
+    aes256_ecb(buffer, ctr, 1, &ctx);
+}
+
+void nist_kat_init(uint8_t *entropy_input, const uint8_t *personalization_string, int security_strength);
```
thomwiggers: is this prototype necessary?

test/crypto_kem/nistkat.c
joostrijneveld: The original KAT format 'addresses' that 'problem' in the sense that it includes secret keys. By requiring consistency with those files, I suppose #93 is forced towards doing the same thing.

thomwiggers: Yeah, Matthias and I talked a bit about it at the coffee machine. The problem isn't limited to secret keys, but also for example the choice of randomness in encapsulation. As identified in #93, I guess that the question indeed is where we draw the line in specifications: do they include these kinds of details?

So I guess the NIST format also doesn't answer any of these questions and any equivalence testing could really only be interoperability of implementations (i.e. do they decapsulate to the same shared secret key).

common/nistkatrng.c
joostrijneveld: Yes, because of `-Wmissing-prototypes` (see #108 and #109)

thomwiggers:
I'm still a bit unclear on how the NIST format works, didn't it have a request/response type of structure? I'm only seeing one side of it here.

dstebila:
> I'm still a bit unclear on how the NIST format works, didn't it have a request/response type of structure? I'm only seeing one side of it here.

I think the request/response structure can be used if you want to write an interactive program that receives the seed inputs, then generates responses, and compare them against the responses.  The way I've implemented here is that it starts from the same master seed, then checks that it generates the same byte-for-byte output as in the response file.

PR #112, opened 2019-04-15
Add remaining schemes from NTRU Round 2 submission

dstebila:
Note that #111 should be merged first since this PR incorporates all of its commits.

* Optional:


dstebila:
The most recent build failed in one of the Windows runs because of a clang-tidy failure.  Why are why even running clang-tidy on Windows?

PR #115, opened 2019-04-16
Fix problems in the NTRU source code

thomwiggers:
Fix issues in the NTRU source code:

* `size_t` and `unsigned int` comparison
* `int` overflows
* `sha3_256` being fed an incorrect length.

### DIFF
--- a/crypto_kem/ntruhps2048509/clean/kem.c
+++ b/crypto_kem/ntruhps2048509/clean/kem.c
@@ -52,7 +52,7 @@ int PQCLEAN_NTRUHPS2048509_CLEAN_crypto_kem_dec(uint8_t *k, const uint8_t *c, co
     for (i = 0; i < NTRU_CIPHERTEXTBYTES; i++) {
         cmp[i] = c[i];
     }
-    sha3_256(rm, cmp, NTRU_PRFKEYBYTES + NTRU_CIPHERTEXTBYTES);
+    sha3_256(rm, cmp, NTRU_CIPHERTEXTBYTES);
 
     PQCLEAN_NTRUHPS2048509_CLEAN_cmov(k, rm, NTRU_SHAREDKEYBYTES, (unsigned char) fail);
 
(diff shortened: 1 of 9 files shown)


crypto_kem/ntruhps2048509/clean/kem.c
```diff
@@ -52,7 +52,7 @@ int PQCLEAN_NTRUHPS2048509_CLEAN_crypto_kem_dec(uint8_t *k, const uint8_t *c, co
     for (i = 0; i < NTRU_CIPHERTEXTBYTES; i++) {
         cmp[i] = c[i];
     }
-    sha3_256(rm, cmp, NTRU_PRFKEYBYTES + NTRU_CIPHERTEXTBYTES);
```
dstebila: This is from the original submission. Are you saying that original code was also wrong?

thomwiggers: Yes, if you see the following:

```c
    uint8_t buf[NTRU_PRFKEYBYTES + NTRU_CIPHERTEXTBYTES];
    uint8_t *cmp = buf + NTRU_PRFKEYBYTES;
```

Then clearly, `cmp + NTRU_PRFKEYBYTES + NTRU_CIPHERTEXTBYTES` is out of bounds.

thomwiggers: I could send an email to the NTRU team.

dstebila: Yes, probably worth checking with them. Very strange if this produces the same KAT values despite the change.

dstebila: > Yes, probably worth checking with them. Very strange if this produces the same KAT values despite the change.

Ah, this branch is only used when decapsulation fails, which doesn't actually happen for the KAT test vectors, so it's not actually affecting KAT values.

thomwiggers:
indeed.

crypto_kem/ntruhps2048509/clean/kem.c
jschanck: Ah, good catch. The length is right but the input should be "buf". Fixed upstream https://github.com/jschanck/ntru/commit/c7fa0b98bcf446077ca3c83c2f9fa81d6bd6f212

Issue #50, opened 2019-02-26
Write guide to using PQClean code in application/library


dstebila:

dstebila:
Fixed by #116.

PR #116, opened 2019-04-16
Write up some lines about some consumers

thomwiggers:
I thought it would be helpful to add a few lines to the README as pointers to ways you could consume source code from PQClean.


dstebila:
Fixes #50.

PR #145, opened 2019-04-17
Add NTRU Prime

### LINKED ISSUE #133: Add NTRU-Prime (Round 3 KEM alternate candidate)

dstebila:
Fixes #133.  

* Optional:


dstebila:
PR deferred at Dan Bernstein's request until revised upstream version released with some functions unrolled and thereby compatible with Microsoft compiler without algorithm changes.

PR #147, opened 2019-04-17
Slow tests

thomwiggers:
Annotate slow tests with the `helpers.slow_test` function decorator, and they'll be skipped if:

* `CI` is set in the environment
* `RUN_SLOW` is not set in the environment.

### DIFF
--- a/test/helpers.py
+++ b/test/helpers.py
@@ -80,6 +80,15 @@ def skip_windows(message="This test is not supported on Windows"):
     return wrapper
 
 
+def slow_test(f):
+    @functools.wraps(f)
+    def wrapper(*args, **kwargs):
+        if 'CI' in os.environ and 'RUN_SLOW' not in os.environ:
+            raise unittest.SkipTest("Slow test skipped on CI run")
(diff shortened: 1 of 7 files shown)


test/helpers.py
```diff
@@ -80,6 +80,15 @@ def skip_windows(*args, **kwargs):
     return wrapper
 
 
+def slow_test(f):
+    @functools.wraps(f)
+    def wrapper(*args, **kwargs):
+        if 'CI' in os.environ and 'RUN_SLOW' not in os.environ:
+            raise unittest.SkipTest("Slow test skipped on CI run")
```
dstebila: It seems to run fast enough in some of the CI environments, so do you think we should turn it for at least one or two of the environments?  Since I develop on macOS I don't run the valgrind tests locally so I would like some assurance for the code that I'm submitting.

thomwiggers: Yeah, it's a tradeoff. Some of the schemes are simply too slow to run on Valgrind though...

dstebila:
Okay. Is this ready to merge?

PR #148, opened 2019-04-17
Add -O3 to CFLAGS

thomwiggers:
This will hopefully speed up tests.


thomwiggers:
I'm just going to yolomerge this, it'll help with the other problems we're having.

PR #152, opened 2019-04-18
Schedules circleci builds with RUN_SLOW

thomwiggers:
Also specifies an order on the PQCLEAN builds to limit fruitless
testing.

Related to #147

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,4 @@
-version: 2
+version: 2.1
 
 .emulated_job: &defaultjob
   machine: true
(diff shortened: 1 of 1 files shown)


joostrijneveld:
Does this mean that subsequent tests are not ran when one test fails? That seems like it may make debugging / fixing errors a bit harder.

thomwiggers:
The other architecture tests will indeed not run when a prior architecture in the priority list fails. I thought that many tests will probably fail on the same problems, and that reducing the number of builds that will be run fruitlessly will allow a newly pushed build to be run sooner instead of being held up by a dozen of queued jobs.

joostrijneveld:
That's probably true.

PR #154, opened 2019-04-23
Detect cron builds on Travis CI

thomwiggers:
Travis doesn't allow setting environment variables on scheduled builds.

I've set up Travis to run a build on master daily, this PR will make sure it handles the slow tests correctly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ matrix:
         - git checkout $TRAVIS_BRANCH
         - git reset --hard $TRAVIS_COMMIT
       script:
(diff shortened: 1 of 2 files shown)


joostrijneveld:
What's the reason behind `travis_wait`? Can you add a comment?

thomwiggers:
Because otherwise tests that take >10m will be terminated.

joostrijneveld:
I'm now noticing that `travis_wait` silences log output while the process is running. If we care about that, we could re-implement using one of the solutions described in https://github.com/travis-ci/travis-ci/issues/4190

PR #160, opened 2019-04-29
Add NewHope

### LINKED ISSUE #131: Add Round 2 KEM NewHope

mkannwischer:
Closes #131 

* Optional:


mkannwischer:
On Windows the functest for the CPA variants seems to fail with `ERROR invalid ciphertext` since even with a few bit flips, it derives the same ss.

See https://github.com/PQClean/PQClean/blob/56a3715ddc15b4647e88530680c44a20433228cd/test/crypto_kem/functest.c#L173-L209
This leads to two questions:
 - Do we want to have CPA parameter sets in PQClean?
 - If yes, should this test be removed?

dstebila:
> * Do we want to have CPA parameter sets in PQClean?

I want to say "no", but BIKE is CPA-only, so we will have at least one CPA scheme.  As for submissions that have a CPA scheme and a CCA scheme, my initial inclination is to be opinionated and only take the CCA, but perhaps we shouldn't be opinionated about what is a good scheme, just about what is a good implementation?

cryptojedi:
> > * Do we want to have CPA parameter sets in PQClean?
 
> I want to say "no", but BIKE is CPA-only, so we will have at least one
> CPA scheme.  As for submissions that have a CPA scheme and a CCA
> scheme, my initial inclination is to be opinionated and only take the
> CCA, but perhaps we shouldn't be opinionated about what is a good
> scheme, just about what is a good implementation?

I would include CPA-secure schemes and then probably change this test
to something more invasive, i.e., try with a completely different,
random, ciphertext. 

If schemes without CCA security should be standarized and what their
value is, what the price for CCA security is etc. is an interesting
discussion, but I think we shouldn't enforce personal preferences for
the outcome of that discussion in PQClean.

dstebila:
We should add a field to META.yml indicating whether the scheme is IND-CCA or IND-CPA.

mkannwischer:
Could someone review this PR? I think it's ready to be merged.

crypto_kem/newhope1024cca/clean/poly.c
```diff
@@ -0,0 +1,370 @@
...
+* Name:        poly_frommsg
+*
+* Description: Convert 32-byte message to polynomial
+*
+* Arguments:   - poly *r:                  pointer to output polynomial
+*              - const unsigned char *msg: pointer to input message
+**************************************************/
+void PQCLEAN_NEWHOPE1024CCA_CLEAN_poly_frommsg(poly *r, const unsigned char *msg) {
+    unsigned int i, j, mask;
+    for (i = 0; i < 32; i++) { // XXX: MACRO for 32
```
dstebila: Should this XXX be resolved?

crypto_kem/newhope1024cca/clean/poly.c
```diff
@@ -0,0 +1,370 @@
...
+*              with output polynomial close to centered binomial distribution
+*              with parameter k=8
+*
+* Arguments:   - poly *r:                   pointer to output polynomial
+*              - const unsigned char *seed: pointer to input seed
+*              - unsigned char nonce:       one-byte input nonce
+**************************************************/
+void PQCLEAN_NEWHOPE1024CCA_CLEAN_poly_sample(poly *r, const unsigned char *seed, unsigned char nonce) {
+    unsigned char buf[128], a, b;
+//  uint32_t t, d, a, b, c;
```
dstebila: Delete this commented line?

crypto_kem/newhope1024cca/clean/poly.c
```diff
@@ -0,0 +1,370 @@
...
+    extseed[NEWHOPE_SYMBYTES] = nonce;
+
+    for (i = 0; i < NEWHOPE_N / 64; i++) { /* Generate noise in blocks of 64 coefficients */
+        extseed[NEWHOPE_SYMBYTES + 1] = (unsigned char) i;
+        shake256(buf, 128, extseed, NEWHOPE_SYMBYTES + 2);
+        for (j = 0; j < 64; j++) {
+            a = buf[2 * j];
+            b = buf[2 * j + 1];
+            r->coeffs[64 * i + j] = hw(a) + NEWHOPE_Q - hw(b);
+            /*
```
dstebila: Deleted this commented-out code block?

crypto_kem/newhope1024cca/clean/poly.c
mkannwischer: This is also in https://github.com/newhopecrypto/newhope/blob/master/ref/poly.c 
@cryptojedi?

mkannwischer:
Thanks @dstebila, I've added the checks and resolved some comments.

cryptojedi:
"Matthias J. Kannwischer" <notifications@github.com> wrote:
> mkannwischer commented on this pull request.

> +void PQCLEAN_NEWHOPE1024CCA_CLEAN_poly_frommsg(poly *r, const unsigned char *msg) {
> +    unsigned int i, j, mask;
> +    for (i = 0; i < 32; i++) { // XXX: MACRO for 32
> 
> This is also in https://github.com/newhopecrypto/newhope/blob/master/ref/poly.c 
> @cryptojedi? 

Could change to NEWHOPE_SYMBYTES.

crypto_kem/newhope1024cca/clean/poly.c
dstebila: Resolved

PR #162, opened 2019-04-30
MQDSS

### LINKED ISSUE #141: Add Round 2 signature scheme MQDSS

joostrijneveld:
This adds both MQDSS parameter sets. Fixes #141.

This also adds a check to see if both Makefiles exist, since it's easy to forget about the Windows Makefile when developing on Linux (and maybe vice versa?). 

* Optional:


test/test_makefiles.py
```diff
@@ -0,0 +1,30 @@
+"""
```
thomwiggers: Maybe rename this to `test_makefiles_present.py`, to reduce confusion with `test_makefile_dependencies`?

dstebila:
Is this ready to merge?

joostrijneveld:
I think so.

PR #165, opened 2019-05-13
Add IND-CPA/IND-CCA2 to metadata

mkannwischer:
This adds IND-CPA/IND-CCA2 to META.yml. 
Currently, all schemes are IND-CCA2, but other schemes are IND-CPA (e.g. NewHope #160) 

Also adapts the metadata test to check if the value is present and contains a valid value.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,6 +25,7 @@ See the section [API](#API) below.
     name: Name
     type: <kem|signature>
     claimed-nist-level: <N>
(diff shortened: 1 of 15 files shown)


joostrijneveld:
Should we revisit #164 based on this?

mkannwischer:
Yes, we could do that, but I'm not sure if it adds a lot of value. 
This came up in #160, so that one can easily see in the META files which schemes only provide IND-CPA security.

PR #170, opened 2019-05-16
Check NIST KAT test vectors for signatures as well

### LINKED ISSUE #169: Check NIST KAT test vectors for signatures as well
Currently, we only check for matching NIST KAT test vectors for KEMs and use our own test vectors for signatures. 
While this is useful to validate that both the NIST signature API and the detached signature API work on all platforms, I think we should still check the NIST KAT test vectors, i.e., do both checks for signatures. 

Is there any reason not to do this? Or does it just need to be implemented?

mkannwischer:
Resolves #169 

* Optional:


joostrijneveld:
I compared these hashes to the KAT files as available on the NIST website, and they indeed match the first test vector.

PR #175, opened 2019-05-20
Use opaque structs for the hashing API

### LINKED ISSUE #171: Incremental hash API constants
The incremental hashing APIs provided by the `common/` directory appear to be doing some assumptions. For example, the incremental hashing API for SHA256 assumes a state of 40 bytes.

Aside from this not being documented, this also leads to some difficulties if anyone wants to substitute the SHA256 implementation. I am trying to replace SHA2 by the Rust Sha2 crate, but because this implementation has a slightly different structure the state is some bytes larger.

I'd propose introducing some constants like `PQCLEAN_SHA2_INCREMENTAL_STATE_BYTES` to then replace the hardcoded constants.

Alternatively, we could also do it with structs, like the AES api.

thomwiggers:
Closes #171

### DIFF
--- a/common/fips202.c
+++ b/common/fips202.c
@@ -520,36 +520,36 @@ static void keccak_inc_squeeze(uint8_t *h, size_t outlen,
     }
 }
 
-void shake128_inc_init(uint64_t *s_inc) {
-    keccak_inc_init(s_inc);
+void shake128_inc_init(shake128incctx *state) {
+    keccak_inc_init(state->ctx);
 }
 
-void shake128_inc_absorb(uint64_t *s_inc, const uint8_t *input, size_t inlen) {
-    keccak_inc_absorb(s_inc, SHAKE128_RATE, input, inlen);
+void shake128_inc_absorb(shake128incctx *state, const uint8_t *input, size_t inlen) {
+    keccak_inc_absorb(state->ctx, SHAKE128_RATE, input, inlen);
 }
 
-void shake128_inc_finalize(uint64_t *s_inc) {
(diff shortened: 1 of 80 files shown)


joostrijneveld:
Looks good to me. Did you run tests for SPHINCS+ locally, given that CI is currently skipping them for efficiency reasons?

thomwiggers:
Yes, on Linux. I'm currently running all tests on `lilo6`.

thomwiggers:
```
2638 tests run in 14273.136 seconds.
55 skipped (2583 tests passed)
```
(clang-tidy was not available)

Seems like this can be merged.

thomwiggers:
Actually: one open question:

Do we want to split the types between absorb-state and squeeze state of a XOF? Because it'd be incorrect form to pass the state back to the absorb function once you've called `finalize`...

thomwiggers:
> Do we want to split the types between absorb-state and squeeze state of a XOF? Because it'd be incorrect form to pass the state back to the absorb function once you've called finalize...

Probably not in C. In e.g. Rust (and probably C++11+) it's easy to do define a function that _consumes_ a value and then returns a new value. In C you'd have to copy the state or do weird things with pointers to pointers... Not sure that's better semantics-wise.

cryptojedi:
Thom Wiggers <notifications@github.com> wrote:
> Actually: one open question:
> 
> Do we want to split the types between absorb-state and squeeze state
> of a XOF? Because it'd be incorrect form to pass the state back to the
> absorb function once you've called `finalize`...

Interesting question. The type system of C is not strong enough to
really enforce this, though. As the state is really the same kind of
object, an implementation can simply cast pointer types and still do
this without triggering even warnings.

PR #176, opened 2019-05-20
Add optimized FrodoKEM

mkannwischer:

* Optional:


mkannwischer:
This is ready for review. The only files changed compared to the clean implementations are `matrix_shake.c` and `matrix_aes.c`.

Issue #157, opened 2019-04-24
Unify Makefiles to reduce compiler flag duplication


joostrijneveld:
In #156, it annoyed me a bit that I had to edit the compiler flags in every Makefile. This also makes it quite likely for these compiler flags to get out of sync when we have more than a handful of schemes.

Can we unify these somehow while still maintaining the 'you can build it as a standalone directory' idea?

I realize we've been over this a few times, and don't intend to re-hash the discussion of whether we need a Makefile, but I feel like the current situation is not optimal.

cryptojedi:
Joost Rijneveld <notifications@github.com> wrote:
> In #156, it annoyed me a bit that I had to edit the compiler flags in
> every Makefile. This also makes it quite likely for these compiler
> flags to get out of sync when we have more than a handful of schemes.
> 
> Can we unify these somehow while still maintaining the 'you can build
> it as a standalone directory' idea?

In the long run we'll probably want to have different flags for
different implementations, at least when starting to support optimized
implementations which may need architecture-specific flags.

How about adding a CI test checking that certain flags (e.g., "-Wall
-Wextra -Wmissing-prototypes) are present?

thomwiggers:
> Can we unify these somehow while still maintaining the 'you can build it as a standalone directory' idea?

I don't think we can really do both, while also just using Makefiles – we would really need something like CMakeLists.txt that supports a more recursive structure (in my understanding) with passing parameters somehow if we want to support both local customisation (needed for AVX2, etc), and stuff from elsewhere.

PR #177, opened 2019-05-20
Add -Wredundant-decls

### LINKED ISSUE #166: Somehow catch redundant declarations
The current SPHINCS code has an issue where the precomputation state array for SHA256 is defined in both `sha256.h` and `sha256.c`. This was caught when trying to link everything together in pqm4, but is not a problem when putting together the `.a` file in pqclean. The difference may be in the fact that pqm4 exposes all .h files through `-I`, but in pqclean it somehow disappears in the indirection created by the `.a`.

I'm not sure if this is a problem we should solve by adding the `-Wredundant-decls` flag to the building of the `.a`, or by changing the build process somewhat (i.e. by linking in the header files).

EDIT: See #167

thomwiggers:
Adds ``-Wredundant-decls`` to close #166.

Also adds some `-O3`s to the NTRU schemes that were missing it.

### DIFF
--- a/crypto_kem/frodokem1344aes/clean/Makefile
+++ b/crypto_kem/frodokem1344aes/clean/Makefile
@@ -4,7 +4,7 @@ LIB=libfrodokem1344aes_clean.a
 HEADERS=api.h params.h common.h
 OBJECTS=kem.o matrix_aes.o noise.o util.o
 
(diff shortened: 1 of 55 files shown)


joostrijneveld:
Shoutout to issue #157.

PR #183, opened 2019-06-07
Add cSHAKE{128,256} to common

### LINKED ISSUE #181: Add support for cSHAKE
This is related to ThreeBears (#127, mupq/pqm4#64). ThreeBears uses cSHAKE with the customization string `"ThreeBears"`. The reference implementation submitted to NIST comes with its own cSHAKE implementation, but for integration into pqm4 and PQClean it would be great if it would share the common Keccak implementation. We would like to add ThreeBears to at least pqm4 within the next few weeks, which is why we would like to settle on a shared API for cSHAKE support under `common/`.

(Note that this is different from the `cshake{128,256}_simple` that some schemes use. That unofficial variant greatly restricts the use of the function-name string and customization string that are defined in NIST SP 800-185. AFAIK ThreeBears is the only scheme with a longer customization string.)

This is what I currently had in mind. Please let me know what you think.

`common/sp800-185.h`:
```c
void cshake128_inc_init(shake128incctx *state, const uint8_t *name, size_t namelen, const uint8_t *cstm, size_t cstmlen);
void cshake128_inc_absorb(shake128incctx *state, const uint8_t *input, size_t inlen);
void cshake128_inc_finalize(shake128incctx *state);
void cshake128_inc_squeeze(uint8_t *output, size_t outlen, shake128incctx *state);
void cshake128(uint8_t *output, size_t outlen, const uint8_t *name, size_t namelen, const uint8_t *cstm, size_t cstmlen, const uint8_t *input, size_t inlen);

void cshake256_inc_init(shake256incctx *state, const uint8_t *name, size_t namelen, const uint8_t *cstm, size_t cstmlen);
void cshake256_inc_absorb(shake256incctx *state, const uint8_t *input, size_t inlen);
void cshake256_inc_finalize(shake256incctx *state);
void cshake256_inc_squeeze(uint8_t *output, size_t outlen, shake256incctx *state);
void cshake256(uint8_t *output, size_t outlen, const uint8_t *name, size_t namelen, const uint8_t* cstm, size_t cstmlen, const uint8_t *input, size_t inlen);
```
cSHAKE is defined to be equal to SHAKE when the function-name string and customization string are empty, so I think it makes sense to reuse the `shake{128,256}incctx` structs.

For comparison, [here](https://github.com/XKCP/XKCP/blob/master/lib/high/Keccak/SP800-185/SP800-185.h) are the prototypes in the XKCP. SP 800-185 can be found [here](https://doi.org/10.6028/NIST.SP.800-185).

Ko-:
This resolves #181. Please let me know what you think.

For the `*_finalize` functions I can't reuse the SHAKE counterparts from `fips202`, because the regular SHAKE uses a different domain separator. I can't reuse `keccak_inc_finalize` because that is not a publicly visible function. That could be changed, but that would require more restructuring and given that it's only three lines of code, I just copied the contents.

The test is based on the test vectors I found [here](https://github.com/XKCP/XKCP/blob/master/tests/UnitTests/testSP800-185.c). The [NIST document](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-185.pdf) doesn't mention any.


mkannwischer:
Thanks, @Ko- !
I've reviewed your changes and everything looks good to me. 
In case all tests pass, I think this can be merged.

Issue #123, opened 2019-04-16
Add Round 2 KEM LEDAcrypt


dstebila:

dstebila:
Only add IND-CCA variants (KEM-LT)

leonbotros:
I'm working on this

dstebila:
Great! Let us know if you have any questions.

mkannwischer:
Still needs to be made constant-time.

PR #186, opened 2019-06-11
LEDAcrypt KEM-LT

leonbotros:
Adds parameter sets of LEDAcrypt KEM-LT that aim for IND-CCA security (issue #123). Currently, it only includes parameters that achieve a DFR of 2<sup>-λ</sup>.
The instances are named ledakemlt{category,n0}, where category is the NIST category and n0 is the number of blocks forming parity-check matrix H (2 for all). The implementation in not constant-time or optimized.

I think commit 6f31e7e should solve problems that arise from casting a pointer to the hash input on big-endian machines. Do we need to serialize the pk and ct as well?

Commit 32b3a97 adds sha3_384 in similar fashion as the others as ledakemlt32 required a digest length of 384 bits.

Feedback would be much appreciated!

* Optional:


thomwiggers:
There's some `.h` files that have implementations in them. Could you move those implementations into a proper `.c` file? See the discussion in #187

thomwiggers:
Travis MacOS PR build failed because of expired timeouts – not sure how to fix that immediately.

The ppc builds on circleci failed because of something else though. The hashes don't match up - this suggests some endianness issue.

leonbotros:
Yes, the ss is correct though. I'm writing some code that should (de)serialize the pk and ct correctly.

thomwiggers:
> Travis MacOS PR build failed because of expired timeouts – not sure how to fix that immediately.

Has been addressed by #190

mkannwischer:
Is this ready for review/merge?

leonbotros:
I have been in contact with the designers regarding solving the global variable (#189),
where Thom pointed out that they should not be used. 

Basically, during key generation it adjusts the decoding threshold for a key pair, which is then used again during decapsulation/decoding. They meant to add this value to the threshold (one uint8_t) to the secret key and pass it as a variable to the decoder. This means these instances' KAT values will have an extra byte in the secret key. I could add this as well now and later check if we have the same KAT values (should be easy), or I could do a PR later.

mkannwischer:
Avoiding global state is currently not a formal requirement for PQClean, is it?
I'd say we go for compatibility with the code submitted to NIST. 

Any other opinions?

dstebila:
> e code submitted to NIST.



> Avoiding global state is currently not a formal requirement for PQClean, is it?
> I'd say we go for compatibility with the code submitted to NIST.

I think we should avoid global state.  We talked before about thread safety being desirable, and global state is incompatible with thread safety. 

And in general if you think about serializing a key at one time / on one computer and deserializing it at another time / on another computer, there's no reason to believe they'll have been able to pass that global state between them.

leonbotros:
It's definitely a bug though. Calling keypair() twice and then decapsulating using the first key pair will result in the threshold for the second key pair being used for decoding. It will not result in a wrong output though as the threshold is lowered for complexity reasons. I have implemented a solution that stores this value in the secret key. I can push it to another branch on my fork and leave it for now.

dstebila:
If this change does result in different KATs, then we shouldn't merge this code unless the Ledacrypt creators state that they truly do intend to make that change in their original specification and KATs.

leonbotros:
LEDAcrypt designers stated that it will be changed in the reference code base and specification.

cryptojedi:
"Matthias J. Kannwischer" <notifications@github.com> wrote:
> Avoiding global state is currently not a formal requirement for PQClean, is it?
> I'd say we go for compatibility with the code submitted to NIST. 

I think we should avoid any global state.

leonbotros:
I'll promise to verify with official KAT values again asap

leonbotros:
Peter and I discussed that we do want to make this a constant-time implementation. Is is ok if I start making such changes in this PR? Can I still tag this as a WIP?

cryptojedi:
leonbotros <notifications@github.com> wrote:
> Peter and I discussed that we do want to make this a constant-time
> implementation. Is is ok if I start making such changes in this PR?
> Can I still tag this as a WIP?

How about getting the non-constant-time version in first (named as
"leaktime" or something) and then work on a constant-time ("clean")
version?

leonbotros:
Then this is ready for review.

crypto_kem/ledakemlt12/leaktime/bf_decoding.c
```diff
@@ -0,0 +1,77 @@
+#include "bf_decoding.h"
+#include "gf2x_arith_mod_xPplusOne.h"
+
+#include <assert.h>
```
mkannwischer: can this be deleted?

crypto_kem/ledakemlt12/leaktime/gf2x_arith.c
```diff
@@ -0,0 +1,76 @@
+#include "gf2x_arith.h"
+
+#include <assert.h>
```
mkannwischer: can we get rid of all the asserts?

crypto_kem/ledakemlt32/leaktime/gf2x_arith_mod_xPplusOne.c
```diff
@@ -0,0 +1,584 @@
...
+    right_bit_shift(NUM_DIGITS_GF2X_ELEMENT, A);
+
+    for (i = NUM_DIGITS_GF2X_ELEMENT - 1; i >= (NUM_DIGITS_GF2X_ELEMENT + 1) / 2; i--) {
+        rev1 = reverse_digit(A[i]);
+        rev2 = reverse_digit(A[NUM_DIGITS_GF2X_ELEMENT - 1 - i]);
+        A[i] = rev2;
+        A[NUM_DIGITS_GF2X_ELEMENT - 1 - i] = rev1;
+    }
+
+    // A[NUM_DIGITS_GF2X_ELEMENT / 2] = reverse_digit(A[NUM_DIGITS_GF2X_ELEMENT / 2]); // no middle digit
```
mkannwischer: delete?

crypto_kem/ledakemlt32/leaktime/gf2x_arith_mod_xPplusOne.c
```diff
@@ -0,0 +1,584 @@
...
+}
+
+/* the implementation is safe even in case A or B alias with the result */
+/* PRE: A and B should be sorted and have INVALID_POS_VALUE at the end */
+void PQCLEAN_LEDAKEMLT32_LEAKTIME_gf2x_mod_add_sparse(
+    int sizeR, POSITION_T Res[],
+    int sizeA, const POSITION_T A[],
+    int sizeB, const POSITION_T B[]) {
+
+    POSITION_T tmpRes[DV * M]; // TODO: now function only works for adding (disjunct) DV and M positions
```
mkannwischer: how about this TODO?

leonbotros: It's only used for adding disjunct locations. I'll add it as a precondition.

thomwiggers:
Matthias said :heavy_check_mark:, the CI builds said :heavy_check_mark:, I say: :ship: it

dstebila:
Where did you get the KAT files for the revised implementation?  I need the full KAT file when adding to liboqs, not just the hash.

leonbotros:
I haven't heard back from them yet. They did tell me they would simply append the uint8_t threshold as a byte to the secret key, so I did just that and hashed it. I also believe I've spotted another bug. Shall I respond here once both have been resolved?

dstebila:
That would be great, thanks.

dstebila:
Any news on updated official KATs?

leonbotros:
Unfortunately, I haven't heard back yet..

PR #188, opened 2019-06-11
Polish Dilithium

mkannwischer:
Found some unnecessary copies and replaced some incremental hashing with a normal hash call. 
Sorry for not noticing this earlier.


thomwiggers:
The build is failing because we now take too much time... I'm going to go ahead and merge this though.

PR #191, opened 2019-06-12
Remove testvectors from test framework

thomwiggers:
They've been superseded by ``test_nistkat``.


mkannwischer:
No, the NIST KATs don't contain the detached signatures. 
See https://github.com/PQClean/PQClean/issues/169
We need both tests for signatures.

PR #192, opened 2019-06-18
Add Saber

mkannwischer:
Fixes #135. 

#### Manually checked properties

* [x] No stringification macros
* [ ] Output-parameter pointers in functions are on the left
* [ ] Negative return values on failure of API functions (within restrictions of FO transform).
* [ ] `const` arguments are labeled as `const`
* [ ] variable declarations at the beginning (except in `for (size_t i=...`)
* Optional:
  * [ ] All integer types are of fixed size, using `stdint.h` types (including `uint8_t` instead of `unsigned char`)
  * [ ] Integers used for indexing are of size `size_t`

PR #195, opened 2019-06-20
Convert `principal-submitter` into a list

### LINKED ISSUE #194: Turn `principal-submitters` into a list.
We should probably turn `principal-submitters` into a list because are some schemes that have more than one principal submitter. We talked about allowing either a list or a string, but now I thought about it a bit more, I don't think we should go in that direction:

* Singleton lists are not that annoying to write or process
* Allowing both `str` and `list` type there is annoying to check
* More importantly: it's very annoying if you are parsing the `META.yml` and need to detect the type of the element instead of just wrapping it in `', '.join(meta['principal-submitters'])`.

Might be best to do that in a separate PR though, because it will touch all schemes.

_Originally posted by @thomwiggers in https://github.com/PQClean/PQClean/pull/192_

thomwiggers:
There are schemes, like SABER (#192) that have more than one principal
submitter. Consistency warrants that we turn it into a list for all
schemes and don't do something with allowing either a str or a list:
that would just be very annoying to parse.

Closes #194

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,8 +33,11 @@ See the section [API](#API) below.
     length-signature: <N>           # Signature only
     nistkat-sha256: sha256sum of 1st NIST KAT test case # KEM and signature
     testvectors-sha256: sha256sum of output of testvectors # Signature only
-    principal-submitter: Eve
-    auxiliary-submitters:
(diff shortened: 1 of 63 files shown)


thomwiggers:
Unsure why appveyor-PR failed...

PR #196, opened 2019-06-20
Update OSX image

thomwiggers:
This solves the problems with the failing travis builds.

PR #198, opened 2019-06-21
Add release function for AES key schedule

dstebila:
liboqs has several AES implementations available to it, one of which (OpenSSL EVP) requires that the key schedule be allocated on the heap and released after use. In particular it cannot be done on the stack (see https://github.com/openssl/openssl/issues/962#issuecomment-208792020). This patch adds a release function for the AES key schedule and uses it in all the places where we currently use AES.  

However, PQClean's AES implementation can be done on the stack, so for PQClean's built-in AES implementation, this is a no-op.  Admittedly, this is mildly ugly.  But I couldn't think of a better way to do it; brilliant suggestions welcome.

See the corresponding OQS patch which depends upon this one. (open-quantum-safe/liboqs#497).

### DIFF
--- a/common/aes.c
+++ b/common/aes.c
@@ -622,3 +622,21 @@ void aes256_ctr(unsigned char *out, size_t outlen, const unsigned char *iv, cons
     aes_ctr(out, outlen, iv, ctx->sk_exp, 14);
 }
 
+#ifdef __GNUC__
+#  define UNUSED(x) UNUSED_ ## x __attribute__((__unused__))
+#else
+#  define UNUSED(x) UNUSED_ ## x
+#endif
+
+void aes128_ctx_release(aes128ctx *UNUSED(r)) {
+    // no-op for PQClean's basic AES operation
+}
+
+void aes192_ctx_release(aes192ctx *UNUSED(r)) {
+    // no-op for PQClean's basic AES operation
+}
+
+void aes256_ctx_release(aes256ctx *UNUSED(r)) {
+    // no-op for PQClean's basic AES operation
+}
+
--- a/common/aes.c
+++ b/common/aes.c
@@ -622,21 +622,18 @@ void aes256_ctr(unsigned char *out, size_t outlen, const unsigned char *iv, cons
     aes_ctr(out, outlen, iv, ctx->sk_exp, 14);
 }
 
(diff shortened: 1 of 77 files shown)


dstebila:
Hmmmm... Windows doesn't like the unused variables in the no-op release function.  gcc and clang support `__attribute__ ((unused))` but that doesn't work on Windows.  Any ideas?

cryptojedi:
> Hmmmm... Windows doesn't like the unused variables in the no-op
> release function.  gcc and clang support `__attribute__ ((unused))`
> but that doesn't work on Windows.  Any ideas?

Can you solve this by using a macro instead of a function?

common/aes.c
```diff
@@ -622,3 +622,21 @@ void aes256_ctr(unsigned char *out, size_t outlen, const unsigned char *iv, cons
...
 }
 
+#ifdef __GNUC__
+#  define UNUSED(x) UNUSED_ ## x __attribute__((__unused__))
+#else
+#  define UNUSED(x) UNUSED_ ## x
+#endif
+
+void aes128_ctx_release(aes128ctx *UNUSED(r)) {
+    // no-op for PQClean's basic AES operation
```
thomwiggers: I'd get rid of the macros and do this instead:
```suggestion
    // no-op for PQClean's basic AES operation
    (void)r;
```

thomwiggers:
This will also need some documentation somewhere (`CONTRIBUTING.md`?). 

Also, don't the FIPS202 or SHA functions have similar issues? If you use OpenSSL EVP to provide hashes, you'll have the same problems.

thomwiggers:
To fix the issues with Travis, please rebase on top of master (see #196).

common/aes.c
dstebila: Thanks, done.

dstebila:
> This will also need some documentation somewhere (`CONTRIBUTING.md`?).

Done.

> Also, don't the FIPS202 or SHA functions have similar issues? If you use OpenSSL EVP to provide hashes, you'll have the same problems.

We don't currently exercise that in OQS, so I haven't encountered it yet.  I was going to treat it as a jump-off-that-bridge-when-we-come-to-it thing.

PR #199, opened 2019-06-21
naming conflicts

elsaVelazquez:
notrandombytes.c  has function randombytes, causing naming conflict

randombytes is defined in multiple places (maybe it's for testing purposes), but it causes errors running code straight out of the box


thomwiggers:
This is intended. You should use **either** `randombytes.c` or `notrandombytes.c`. We need the different implementations to provide not-random bytes for our test vectors. This relies exactly on them having the same name.

thomwiggers:
I'm writing a PR to move these test support files into the `test` subdirectory.

PR #201, opened 2019-06-21
Eliminate the global state

### LINKED ISSUE #168: Thread-safety of SPHINCS+-SHA256/Haraka
The SHA2 and the Haraka variants of SPHINCS+ are not thread-safe due to global state.

This is a significant problem for downstream consumers, including [liboqs](https://github.com/open-quantum-safe/liboqs/pull/483) and [rustpq](https://github.com/rustpq/pqcrypto).

thomwiggers:
This PR sacrifices passing some extra arguments to get rid of the global
state. Closes #168.

* Haraka needs state in all hash calls, this results in changes to the
  hash functions specified in `hash.h`.  The extra pointers passed would
  not be necessary for SHA256 or SHAKE256.
* SHAKE256 did not have global state, but uniformity in the implementations
  requires us to pass  around the new state context anyway. Otherwise,
  @joostrijneveld's SPHINCS+ generator doesn't really work anymore.

We introduce a new header file called `primitive.h` which defines the
required state type for the generic functions. I did not go into
replacing _all_ occurrences of state variables by the new `hash_state`
macro.

* Optional:


thomwiggers:
Note that *tests are not run on the CI environment* because SPHINCS+ is too slow. I am currently running tests elsewhere.

thomwiggers:
This branch now contains a possible fix to the unreliable Travis CI PR tests (where it merges the branch into master). If successful, I will factor it out into a different PR.

Issue #122, opened 2019-04-16
Add Round 2 KEM HQC


dstebila:

dstebila:
I've emailed the HQC team and they're interested, they'll follow up with us next week, and then I'll have a clearer idea of how this one will proceed.

mkannwischer:
Resolved by #202

PR #202, opened 2019-06-24
HQC submission

sebastianv89:
HQC, cleaned up version of NIST Round 2 submission (reference implementation). See also issue #122. Some notable points:
* Replaced NTL dependence with my own code for GF(2)[x] multiplication.
* Added NIST seed expander to common files. It works, but I'm not sure this is what we want?
* Code is not constant-time (depends on data). Constant time decoding is a non-trivial problem.
* Missing LICENSE (no license was submitted to NIST, I will send out an email to the HQC team)

Once this version works and can be merged, I will add other HQC 

The following should be satisfied (except indexing is not always done with `size_t`):

* Optional:


common/nistseedexpander.h
```diff
@@ -0,0 +1,37 @@
+#ifndef NISTSEEDEXPANDER_H 
+#define NISTSEEDEXPANDER_H
+
+//
+//  rng.h
+//
+//  Created by Bassham, Lawrence E (Fed) on 8/29/17.
+//  Copyright © 2017 Bassham, Lawrence E (Fed). All rights reserved.
```
thomwiggers: Are we allowed to redistribute this?

sebastianv89: We already have it in nistkatrng.c (https://github.com/PQClean/PQClean/blob/be941b28a8d568378f922d9aac7db0e2ca73d182/common/nistkatrng.c#L5)
I think it even makes sense to merge these two files so they are closer to the original rng.h distributed by NIST.

thomwiggers:
Is there any progress with getting the license files? Also, it appears this needs to get rebased...

common/randombytes.h
```diff
@@ -8,6 +8,6 @@
 #include <unistd.h>
 #endif
 
-int randombytes(uint8_t *buf, size_t xlen);
+int randombytes(uint8_t *buf, size_t n);
```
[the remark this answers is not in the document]
sebastianv89: Just so its the same as randombytes.c

sebastianv89:
> Is there any progress with getting the license files? Also, it appears this needs to get rebased...

Not yet, I've sent a reminder to the HQC team.

common/nistseedexpander.h
sebastianv89: I'm not sure I understand the problem. The `nistseedexpander.h` and `.c` file define the `seedexpander` function that was provided by NIST. HQC does not rely on `nistkatrng.h` at all, but instead just uses `randombytes.h`.

mkannwischer:
What's the status on this? Can we close or merge this?

sebastianv89:
I think it can be closed. The HQC team told me: "We plan to create a merge request on PQClean next week that will provide both reference and optimized implementations."
That was three weeks ago...

sebastianv89:
The HQC team created a pull-request to my branch. I merged it, cleaned it up a bit, and now I'm opening up the pull request here to see if it passes the tests.

crypto_kem/hqc-128-1-cca2/avx2_leaktime/gf2x.c
```diff
@@ -0,0 +1,144 @@
...
+#include "parameters.h"
+#include <stdint.h>
+#include <string.h>
+
+#define VECTORSIZE 256 /*!< The size in bits of a vector*/
+#define BYTESPERVECTOR (VECTORSIZE/8) /*!< The number of bytes in a vector*/
+#define BITSPERINT 32 /*!< The number of bits per int*/
+#define INTSPERVECTOR (VECTORSIZE/BITSPERINT) /*!< The number of ints per vector*/
+#define VEC_N_ARRAY_SIZE_VEC CEIL_DIVIDE(PARAM_N,256) /*!< The number of needed vectors to store PARAM_N bits*/
+typedef uint32_t v8uint32 __attribute__ ((vector_size (BYTESPERVECTOR), aligned (BYTESPERVECTOR)));/*!< v8uint32: a vector of 8 unsigned ints*/
```
thomwiggers: ``__attribute__`` is a GNU compiler extension, and not available on Windows. You should use other tricks to enforce alignment – but this particular example should just use ``_m256i`` from ``immintrin.h``.

crypto_kem/hqc-128-1-cca2/avx2_leaktime/fft.c
```diff
@@ -0,0 +1,631 @@
...
+ * @param[out] f Array receiving the syndromes
+ * @param[in] w Array storing the family
+ * @param[in] f_coeffs Length of syndromes vector
+ * @param[in] m 2^m is the smallest power of 2 greater or equal to the length of family w
+ * @param[in] m_f 2^{m_f} is the smallest power of 2 greater or equal to the length of f
+ * @param[in] betas FFT constants
+ */
+static void fft_t_rec(uint16_t *f, const uint16_t *w, size_t f_coeffs,
+                      uint32_t m, uint32_t m_f, const uint16_t *betas) {
+    size_t k = 1 << (m - 1);
```
thomwiggers: There is a `clang-tidy` warning on this line.

crypto_kem/hqc-128-1-cca2/avx2_leaktime/gf2x.c
sebastianv89: I removed the avx2 implementation.

crypto_kem/hqc-128-1-cca2/avx2_leaktime/fft.c
sebastianv89: I removed the avx2 implementation.

thomwiggers:
There appear to be issues with 32-bit platforms, but maybe this is something we could take over if you don’t have much time to spend on this.

sebastianv89:
I would appreciate that. Let me know if I can be of any help.

thomwiggers:
I ran the tests with EXTRAFLAGS="-m32" and solved a few of the 32-bit problems that came up. I hope it's enough, otherwise CI will tell us soon enough.

thomwiggers:
Celebrated too early – running it in the container shows `clang-tidy` also has a bunch of feedback.

thomwiggers:
There was undetected undefined behaviour in `gf2x.c` where some of the shifts could be larger than `WORD_SIZE_BITS`, ie. larger than the width of ``uint64_t``. This apparently on Intel gets interpreted as the shift mod 64, but on ARM something else happened.

thomwiggers:
@mkannwischer did you still want to look at this?

Issue #197, opened 2019-06-20
LEDA is too slow for the 'slow' valgrind tests


thomwiggers:
We run the 'slow' tests on a schedule, and LEDA appears to break them because it takes too long under valgrind: https://circleci.com/gh/PQClean/PQClean/4983

leonbotros:
I am aware of this problem, see #193.

thomwiggers:
This got fixed, as #206 included a `no_output_timeout: 2h` change.

leonbotros:
Nice job!

I'm also working on speeding this up using faster multiplication methods without variable-sized arrays. I was thinking about including this in a (hopefully more) constant-time "clean" implementation.

PR #206, opened 2019-07-17
Parallel tests

thomwiggers:
This is a big list of changes, but the result should be worth it. I migrated the unit tests from `nosetest` to `pytest`. `pytest` is slightly harder to write generated tests for, but it's still maintained, much easier to use (proper filters!) and most importantly: it allows to run tests in parallel.

To run tests in parallel, I'm introducing an isolation mechanism. This yet again complicates the tests, but I think it's worth it for the speed gains. Running slow tests in parallel saves many, many minutes.

For reference, running all tests on `lilo6`, a 32-threads (16 cores) Xeon Silver 4110 system, used to take several hours. This has now been reduced to ~20 minutes (results pending), excluding the runtime of the slowest valgrind test (LEDAcryptKEM-LT52), which ran over 45 minutes before I killed it. If you run `pytest -n auto -k "not valgrind and not sanitizers"` then it's done in 5:30m.

The filters are also nice: you can now for example do:

* `pytest -k "not sphincs"`
* `pytest -k "kyber and not kyber512"`

etc.

See also #197, #147.

### DIFF
--- a/test/test_char.py
+++ b/test/test_char.py
@@ -9,15 +9,12 @@ import pycparser
 import os
 import helpers
 
+import pytest
 
-def test_char():
+
+def setup_module():
     if not(os.path.exists(os.path.join('pycparser', '.git'))):
-        helpers.run_subprocess(
-            ['git', 'submodule', 'update', '--init']
-        )
-    for scheme in pqclean.Scheme.all_schemes():
-        for implementation in scheme.implementations:
-            yield check_char, implementation
+        print("Please run `git submodule update --init`")
 
 
 def walk_tree(ast):
@@ -29,9 +26,14 @@ def walk_tree(ast):
         yield from walk_tree(child)  # recursively yield prohibited nodes
 
 
+@pytest.mark.parametrize(
+    'implementation',
+    pqclean.Scheme.all_implementations(),
+    ids=str,
+)
 @helpers.filtered_test
 @helpers.skip_windows()
-def check_char(implementation):
+def test_char(implementation):
     errors = []
     for fname in os.listdir(implementation.path()):
         if not fname.endswith(".c"):
@@ -61,12 +63,3 @@ def check_char(implementation):
             "Prohibited use of char without explicit signed/unsigned" +
             "".join(errors)
         )
-
-
-if __name__ == '__main__':
-    try:
-        import nose2
-        nose2.main()
-    except ImportError:
-        import nose
-        nose.runmodule()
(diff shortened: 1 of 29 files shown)


test/test_char.py
```diff
@@ -61,12 +63,3 @@ def check_char(implementation):
             "Prohibited use of char without explicit signed/unsigned" +
             "".join(errors)
         )
-
-
-if __name__ == '__main__':
```
mkannwischer: shouldn't this be replaced with pytest instead of removed?

test/test_linter.py
```diff
@@ -45,10 +45,8 @@ def check_tidy(implementation: pqclean.Implementation):
...
-        sys.argv = sys.argv[0:1] + sys.argv[2:]
-    try:
-        import nose2
-        nose2.main()
-    except ImportError:
-        import nose
-        nose.runmodule()
+        del sys.argv[1]
+
+    import pytest
```
mkannwischer: shouldn't this be in an `if __name__ == '__main__':` block?

test/test_symbol_namespace.py
```diff
@@ -46,13 +49,11 @@ def check_symbol_namespace(implementation):
...
 
 
 if __name__ == '__main__':
-    try:
-        import nose2
-        nose2.main()
-    except ImportError:
-        import nose
-        nose.runmodule()
+    import pytest
```
mkannwischer: pytest is already included at top of file, right?

test/test_functest.py
```diff
@@ -5,87 +5,87 @@
...
         raise unittest.SkipTest("Clang does not support ASAN on ppc")
     elif platform.machine() in ['armv7l', 'aarch64']:
         env = {'ASAN_OPTIONS': 'detect_leaks=0'}
     elif platform.system() == 'Darwin':
-        raise unittest.SkipTest('valgrind is not reliable on OSX')
+        raise unittest.SkipTest('ASAN is not reliable on OSX')
     else:
         print("Supported platform: {}".format(platform.machine()))
 
+    initializer()
```
mkannwischer: rename `init()` as in other tests?
Also should this be called in all tests or is it only required for some?

test/test_functest.py
```diff
@@ -5,87 +5,87 @@
...
-                     working_dir=os.path.join('..', 'test'))
+    helpers.run_subprocess(
+        [os.path.join(dest_dir, 'functest_{}_{}{}'.format(
+            implementation.scheme.name,
+            implementation.name,
+            '.exe' if os.name == 'nt' else ''
+        ))],
+        env=env,
+    )
+    destructor()
```
mkannwischer: rename to `destr()`?

PR #210, opened 2019-07-21
Falcon implementations (integer-only code, constant-time).

pornin:
This is a portable, constant-time, and hopefully clean enough implementation of Falcon (Falcon-512 and Falcon-1024). It uses only integer types, for optimal portability; internally, it contains constant-time emulation of IEEE-754 floating-point operations on double-precision values.

This is an excerpt of what should soon become the new official Falcon code (which also contains versions that leverage a hardware FPU, AVX2 opcodes, or ARM Cortex-M4 assembly -- these extra versions have been removed from this submission). Compared to the "round 2" Falcon reference version, apart from the use of integer types, the Gaussian sampler has been modified, so that observation of the rejection rate does not yield statistically usable information on the secret key (i.e. to allow the code to be called "constant-time"). A consequence is that test vectors have changed (from the same seed, we do not get the same keys and signatures, though produced signatures are still interoperable); also, a new encoding format has been used for private keys, making them substantially shorter. I have set the hash values in the META.yml files correspondingly.

Tests pass on my machine, except "test_char" (it emits "AttributeError: module 'pycparser' has no attribute 'parse_file'" and I don't think it's my fault), and "test_linter". Clang-tidy emits lots of things, including some errors on files which are in common/, thus nominally outside of my reach.

I have included a Makefile for MSVC / nmake, but I have not tested it at all.

The code has been tested successfully on 32-bit and 64-bit machines, both little-endian (x86, ARM) and big-endian (ppc).

* Optional:


crypto_sign/falcon-1024/META.yml
```diff
@@ -0,0 +1,23 @@
...
+  - Paul Kirchner
+  - Vadim Lyubashevsky
+  - Thomas Pornin
+  - Thomas Ricosset
+  - Gregor Seiler
+  - William Whyte
+  - Zhenfei Zhang
+implementations:
+    - name: clean
+      version: https://github.com/FIX-THIS/YES-BUT-HOW/DUNNO-GUV
```
thomwiggers: :thinking: 

But in all seriousness, 'round two' or whatever should also be fine.

crypto_sign/falcon-1024/clean/Makefile.Microsoft_nmake
```diff
@@ -0,0 +1,23 @@
+# This Makefile can be used with Microsoft Visual Studio's nmake using the command:
+#    nmake /f Makefile.Microsoft_nmake
+
+LIBRARY=libfalcon-1024_clean.lib
+OBJECTS=codec.obj common.obj fft.obj fpr.obj keygen.obj pqclean.obj rng.obj sign.obj vrfy.obj
+
+# Warning C4146 is raised when a unary minus operator is applied to an
+# unsigned type; this has nonetheless been standard and portable for as
+# long as there has been a C standard, and we do that a lot, especially
+# for constant-time computations. Thus, we disable that spurious warning.
```
thomwiggers: I'm not entirely sure about this, but I also don't exactly know how trivial this warning is. As it worked everywhere, I'm inclined to leave this be.

thomwiggers:
> Clang-tidy emits lots of things, including some errors on files which are in common/, thus nominally outside of my reach.

This might be due to different versions of Clang – unfortunately clang-tidy is not extremely stable in its warnings.

crypto_sign/falcon-1024/META.yml
pornin: The perils of doing things too late... I changed the version for "round two".

crypto_sign/falcon-1024/clean/Makefile.Microsoft_nmake
pornin: There is some historical context in [Microsoft docs](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4146?view=vs-2019): the main reason for this warning is for people writing "-2147483648" and expecting a signed integer; instead, the constant (without the "-") is too large to fit an "int", and instead has type "unsigned", and the mathematical result of the negation will be an "unsigned" of value +2147483648 instead.

Now, the Microsoft explanation is actually wrong: in C99, or even in C90 on 64-bit systems, "2147483648" as a literal constant will not fit in an "int", so it will be promoted to "long" or "long long" (whichever is large enough), so still signed. Only in C90 (no "long long") on a 32-bit system (so a 32-bit "long") will it become "unsigned". But, at least, this explains what the Microsoft engineers were thinking about.

In many places in the Falcon code, I use negation on unsigned types to get the standard effect, which involves reduction modulo the size of the type (this is specified in the C standard, section 6.2.5/9: "A computation involving unsigned operands can never overflow, because a result that cannot be represented by the resulting unsigned integer type is reduced modulo the number that is one greater than the largest value that can be represented by the resulting type."). Most of the time, this is to transform a bit result (value 0 or 1) into a full-width mask (0x00000000 or 0xFFFFFFFF).

To avoid the C4146 warning, I would have to replace all cases of `-x` with `0-x`, which is cumbersome (it would require extra parentheses) and unsatisfying: I want the compiler to use the `neg` opcode, not a `sub`, and it will do that only if it notices that the subtraction is really a negation in disguise; but, to avoid the warning, I need the compiler _not_ to notice it.

I usually disable the warning with a `#pragma`:

    #if defined _MSC_VER && _MSC_VER
    #pragma warning( disable : 4146 )
    #endif

but then, there will be a `#if` and it will fail `test_preprocessor.py`. Hence the extra flag for the compiler.

thomwiggers:
If you want, it might be interesting to also add the hardfloat version to PQClean – AFAIK all our current targets support floats just fine and we do have support for having parallel implementations (see the FrodoKEM folders). That version could be called `hardfloat`, `clean-float` or something along those lines.

crypto_sign/falcon-1024/clean/Makefile.Microsoft_nmake
thomwiggers: Great explanation.

The Microsoft explanation probably stems from the fact that they don't support C90, and C99 also not really nicely (only the parts needed to support C++ versions).

cryptojedi:
Thomas Pornin <notifications@github.com> wrote:
> pornin commented on this pull request.
> 
> 
> 
> > @@ -0,0 +1,23 @@
> +# This Makefile can be used with Microsoft Visual Studio's nmake using the command:
> +#    nmake /f Makefile.Microsoft_nmake
> +
> +LIBRARY=libfalcon-1024_clean.lib
> +OBJECTS=codec.obj common.obj fft.obj fpr.obj keygen.obj pqclean.obj rng.obj sign.obj vrfy.obj
> +
> +# Warning C4146 is raised when a unary minus operator is applied to an
> +# unsigned type; this has nonetheless been standard and portable for as
> +# long as there has been a C standard, and we do that a lot, especially
> +# for constant-time computations. Thus, we disable that spurious warning.
> 
> There is some historical context in [Microsoft docs](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4146?view=vs-2019): the main reason for this warning is for people writing "-2147483648" and expecting a signed integer; instead, the constant (without the "-") is too large to fit an "int", and instead has type "unsigned", and the mathematical result of the negation will be an "unsigned" of value +2147483648 instead.
> 
> Now, the Microsoft explanation is actually wrong: in C99, or even in C90 on 64-bit systems, "2147483648" as a literal constant will not fit in an "int", so it will be promoted to "long" or "long long" (whichever is large enough), so still signed. Only in C90 (no "long long") on a 32-bit system (so a 32-bit "long") will it become "unsigned". But, at least, this explains what the Microsoft engineers were thinking about.
> 
> In many places in the Falcon code, I use negation on unsigned types to get the standard effect, which involves reduction modulo the size of the type (this is specified in the C standard, section 6.2.5/9: "A computation involving unsigned operands can never overflow, because a result that cannot be represented by the resulting unsigned integer type is reduced modulo the number that is one greater than the largest value that can be represented by the resulting type."). Most of the time, this is to transform a bit result (value 0 or 1) into a full-width mask (0x00000000 or 0xFFFFFFFF).
> 
> To avoid the C4146 warning, I would have to replace all cases of `-x` with `0-x`, which is cumbersome (it would require extra parentheses) and unsatisfying: I want the compiler to use the `neg` opcode, not a `sub`, and it will do that only if it notices that the subtraction is really a negation in disguise; but, to avoid the warning, I need the compiler _not_ to notice it.
> 
> I usually disable the warning with a `#pragma`:
> 
>     #if defined _MSC_VER && _MSC_VER
>     #pragma warning( disable : 4146 )
>     #endif
> 
> but then, there will be a `#if` and it will fail `test_preprocessor.py`. Hence the extra flag for the compiler.

I fully agree.

I would suggest to disable this warning globally in the Microsoft
Makefiles.

thomwiggers:
> I would suggest to disable this warning globally in the Microsoft Makefiles.

This is a discussion that doesn't really concern this PR, but anyway:

I would not do that. This is probably a fairly rare problem, and then you would have this weird, seemingly nonsensical flag in the NMakefiles for all schemes. We could however write a small note about this warning and point to Thomas' excellent description on the Wiki or in CONTRIBUTING.md though.

pornin:
@thomwiggers About adding a "hardfloat" implementation, it's a bit tricky. At some points I need to do round-to-nearest-integer operations; the rint() function from libm is not constant-time, so instead I do that with: (x + 2**52) - 2**52 (for a value x between 0 and 2**52). This trick works great as long as the underlying platform uses exact IEEE-754 rules on double-precision values (52 mantissa bits, exactly). In particular, this fails on 32-bit x86, where the 387 FPU is used and all computations are done internally on an extended type (80 bits, including 64 mantissa bits). The result is that the signature process computes wrong, finds a vector which is not short enough, and loops forever...

This can be fixed by forcing the compiler to use SSE2 instead (`-msse2 -mfpmath=sse`) (apparently this is the default on 32-bit Darwin, but not on Linux) or by reducing the FPU precision (bits 8 and 9 of the 387 FPU control world) (apparently this is the default on 32-bit FreeBSD, but, again, not on Linux). This is hardly "clean", though.

I could also use a `union` between `double` and `uint64_t` and then make the rounding with pure (constant-time) integer operations, but this requires making assumptions on how floating-point types are being represented in memory (that one would be decently portable in practice, since even with the 387 FPU, the in-memory storage is rounded down to the 52 bits of mantissa).

Basically, I don't know how to make a constant-time rint() function that does not make platform-specific assumptions about precision or representation of floating-point types.

thomwiggers:
Ah, that's quite interesting. Yeah, in that case that would need to hold off to until such a time when we want to a) accept slightly-less portable implementations and b) support specifying exactly what targets an implementation supports. Thanks for the explanation!

thomwiggers:
With `-Wcast-align=strict`, the GCC compiler complains about one line in `common.c`, see https://github.com/PQClean/PQClean/pull/221 where we are discussing that warning.

PR #211, opened 2019-07-22
Fix documentation for Dilithium `crypto_sign_open`

### LINKED ISSUE #205: crypto-sign arguments in Dilithium
To correctly return 0 for computing signed message, Line 307 in sign.c, Dilithium 3, explanation should read:
- const unsigned char *sk: pointer to bit-packed secret key
(vs to public key)

thomwiggers:
It mentioned `sk` instead of `pk` as specified in the arguments.

Closes #205.


mkannwischer:
Looks good, I've also pushed that upstream: https://github.com/pq-crystals/dilithium/pull/8

PR #213, opened 2019-07-24
Only absorb root part of pk in haraka message hashing.

kste:
Fix to only absorb root part of the pk in the message hash of Haraka, as required in the NIST specification.

* Optional:

### DIFF
--- a/crypto_sign/sphincs-haraka-128f-robust/META.yml
+++ b/crypto_sign/sphincs-haraka-128f-robust/META.yml
@@ -4,8 +4,8 @@ claimed-nist-level: 1
 length-public-key: 32
 length-secret-key: 64
 length-signature: 16976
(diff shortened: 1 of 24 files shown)


thomwiggers:
Appveyor is probably just being weird.

thomwiggers:
Tests are also not run for SPHINCS+ at the moment (once #206 gets merged we might include it again) – so we should make sure to tests this locally.

dstebila:
> This changes the KAT and testvector values, so the META files should be updated.

Why are the KAT values changing?  Did the specification change, or did our implementation not match the specification?

thomwiggers:
Apparently the SPHINCS+ team made a small mistake. They also made this change in their canonical repository: https://github.com/sphincs/sphincsplus/commit/0208952716af0e023e91f2a64668ce125aa40090

thomwiggers:
I've pushed new hashes; will now run tests on a server somewhere. Should be able to collect the results and press the merge button tomorrow.

PR #214, opened 2019-07-29
Run SPHINCS+ tests again

thomwiggers:
Since #206 got merged, this should be possible again.


thomwiggers:
Hmm we exceeded the time limits at appveyor :cry:

thomwiggers:
this does not seem possible after all, for now. :cry:

Issue #220, opened 2019-08-01
Rainbow has undefined behaviour


thomwiggers:
`blas_u32.c` contains casts of `uint8_t*` to `uint32_t*`, which isn't defined behaviour if the pointer is not aligned – and I'm getting crashes in `rainbowIIIc-cyclic-compressed`. 

A solution would probably be to use `uint32_t` pointers everywhere... If that's possible.

thomwiggers:
This is caught by `test_functest_sanitizers` _sometimes_. It can only tell you when it's going wrong that we have this problem, but not if it's going fine, unfortunately.

thomwiggers:
`-Wcast-align=strict` does catch this, but it is not supported by GCC 7.4

thomwiggers:
I also get stack overflows...

PR #221, opened 2019-08-01
Run with ``-Wcast-align=strict`` when available

thomwiggers:
This warning is not available on Clang or GCC < 8, but it's useful to prevent
problems such as those found in #220.

### DIFF
--- a/test/helpers.py
+++ b/test/helpers.py
@@ -2,6 +2,7 @@ import atexit
 import functools
 import logging
 import os
+import re
 import shutil
 import subprocess
 import sys
@@ -120,9 +121,13 @@ def make(*args, working_dir='.', env=None, expected_returncode=0, **kwargs):
         for envvar in ['IMPLEMENTATION', 'SCHEME']:
             if envvar in kwargs:
                 kwargs['{}_UPPERCASE'.format(envvar)] = (
-                        kwargs[envvar].upper().replace('-', ''))
+                    kwargs[envvar].upper().replace('-', ''))
     else:
         make_command = ['make']
+
+    if supports_cast_align_strict():
+        kwargs['EXTRAFLAGS'] = (
+            kwargs.get('EXTRAFLAGS', '') + ' -Wcast-align=strict')
 
     return run_subprocess(
         [
@@ -272,3 +277,22 @@ def filtered_test(func):
         else:
             raise unittest.SkipTest("Test disabled by filter")
     return wrapper
+
+
+def supports_cast_align_strict():
+    """Tells us if this version of GCC supports -Wcast-align=strict
+
+    This is true if gcc >= 8
+    """
+    if os.name == 'nt':
+        return False
+    try:
+        gcc_version = subprocess.check_output(
+            ['gcc', '--version']).split(b'\n')[0]
+        version_number = re.match(r'gcc \(.*\) (\d)\.\d+\.\d+$',
+                                  gcc_version.decode())
+        if version_number:
+            return int(version_number.group(1)) >= 8
+    except subprocess.CalledProcessError:
+        pass
+    return False
--- a/test/test_functest.py
+++ b/test/test_functest.py
@@ -24,7 +24,7 @@ def test_functest(implementation, impl_path, test_dir,
                   init, destr):
     init()
     dest_dir = os.path.join(test_dir, 'bin')
-    helpers.make('functest',
+    helpers.make('clean-scheme', 'functest',
                  TYPE=implementation.scheme.type,
                  SCHEME=implementation.scheme.name,
                  IMPLEMENTATION=implementation.name,


thomwiggers:
This finds problems in:

* Falcon-512, 1024: 
```c
void NS_hash_to_point(..., uint8_t *tmp) {
// [snip]
tt1 = (uint16_t*) tmp;
```
* FrodoKEM:
```c
kem.c: In function ‘PQCLEAN_FRODOKEM1344AES_OPT_crypto_kem_enc’:                                                                                                                                                                                                                                                              
kem.c:121:47: error: cast increases required alignment of target type [-Werror=cast-align]                                                                                                                                                                                                                                    
  121 |     PQCLEAN_FRODOKEM1344AES_OPT_key_encode(C, (uint16_t *)mu);                                                                                                                                                                                                                                                        
      |                                               ^                                                                                                                                                                                                                                                                       
kem.c: In function ‘PQCLEAN_FRODOKEM1344AES_OPT_crypto_kem_dec’:                                                                                                                                                                                                                                                              
kem.c:158:28: error: cast increases required alignment of target type [-Werror=cast-align]                                                                     
  158 |     const uint16_t *sk_S = (uint16_t *) &sk[CRYPTO_BYTES + CRYPTO_PUBLICKEYBYTES];                                                                     
      |                            ^                                                                                                                                                                                                                                                                                          
kem.c:183:44: error: cast increases required alignment of target type [-Werror=cast-align]                                                                                                                                                                                                                                    
  183 |     PQCLEAN_FRODOKEM1344AES_OPT_key_decode((uint16_t *)muprime, W);                                                                                                                                                                                                                                                   
      |                                            ^                                                                                                                                                                                                                                                                          
kem.c:206:48: error: cast increases required alignment of target type [-Werror=cast-align]                                                                                                                                                                                                                                    
  206 |     PQCLEAN_FRODOKEM1344AES_OPT_key_encode(CC, (uint16_t *)muprime);                                                                                                                                                                                                                                                  
      |                                                ^                                                        
```

* RAINBOW
```c
blas_u32.c: In function ‘PQCLEAN_RAINBOWIACYCLICCOMPRESSED_CLEAN_gf256v_predicated_add_u32’:
blas_u32.c:9:23: error: cast increases required alignment of target type [-Werror=cast-align]
    9 |     uint32_t *b_u32 = (uint32_t *)accu_b;
      |                       ^                                                
blas_u32.c:10:29: error: cast increases required alignment of target type [-Werror=cast-align]
   10 |     const uint32_t *a_u32 = (const uint32_t *)a;
      |                             ^                                          
blas_u32.c: In function ‘PQCLEAN_RAINBOWIACYCLICCOMPRESSED_CLEAN_gf256v_add_u32’:
blas_u32.c:25:23: error: cast increases required alignment of target type [-Werror=cast-align]
   25 |     uint32_t *b_u32 = (uint32_t *)accu_b;
      |                       ^                                                
blas_u32.c:26:29: error: cast increases required alignment of target type [-Werror=cast-align]
   26 |     const uint32_t *a_u32 = (const uint32_t *)a;
      |                             ^                                          
blas_u32.c: In function ‘PQCLEAN_RAINBOWIACYCLICCOMPRESSED_CLEAN_gf16v_mul_scalar_u32’:
blas_u32.c:42:23: error: cast increases required alignment of target type [-Werror=cast-align]
   42 |     uint32_t *a_u32 = (uint32_t *)a;
      |                       ^                                                
blas_u32.c: In function ‘PQCLEAN_RAINBOWIACYCLICCOMPRESSED_CLEAN_gf16v_madd_u32’:
blas_u32.c:65:23: error: cast increases required alignment of target type [-Werror=cast-align]
   65 |     uint32_t *c_u32 = (uint32_t *)accu_c;
      |                       ^                                                
blas_u32.c:66:29: error: cast increases required alignment of target type [-Werror=cast-align]
   66 |     const uint32_t *a_u32 = (const uint32_t *)a;
      |                             ^                                          
blas_u32.c: In function ‘PQCLEAN_RAINBOWIACYCLICCOMPRESSED_CLEAN_gf16v_dot_u32’:
blas_u32.c:90:29: error: cast increases required alignment of target type [-Werror=cast-align]
   90 |     const uint32_t *a_u32 = (const uint32_t *)a;
      |                             ^                                          
blas_u32.c:91:29: error: cast increases required alignment of target type [-Werror=cast-align]
   91 |     const uint32_t *b_u32 = (const uint32_t *)b;
      |                             ^                                          
cc1: all warnings being treated as errors
```

thomwiggers:
So do we want to enable this extra strict warning, which would have prevented #220, or is this complaining about upcasting `uint8_t` to `uint16_t` too strict?

pornin:
The Falcon _internal_ functions expect temporary buffers to be passed as `uint8_t*` but to be suitably aligned for whatever they are doing with it. Depending on the function, data stored in it will consist in bytes, 16-bit, 32-bit and/or 64-bit words. Some functions need to store a combination of those types, but alignment is preserved thanks to the (standard-guaranteed) fact that exact-width types have no padding bits, thus the size of `uint64_t` is necessarily equal to 8 times the size of `uint8_t`.

Callers use unions to allocate suitably-aligned buffers; e.g.: https://github.com/PQClean/PQClean/blob/6e12692b4e8484340779fc90676c6af5e3d8f039/crypto_sign/falcon-512/clean/pqclean.c#L46

It is quite understandable that a compiler would not be able to automatically follow all parameters and decide that all accesses are, in fact, properly aligned. In some cases, this hinges on the fact that the Falcon degree is (in internal functions) a power of two from 2 to 1024, and never equal to 1. I would expect the warning to complain in many places, in fact, not just one; maybe the compilation just stopped on that file and did not go further? When it reaches `keygen.c`, the `-Wcast-align=strict` should light up like a Christmas tree.

If this warning is to be activated then I can probably patch all the code to use unions in declarations or something like that, but it is not going to be pretty.

Alignment is a serious issue. Instead of (or in addition to) the warning, I would suggest running the code on a system that does not tolerate unaligned accesses. Unfortunately, these have been rarefying, and QEMU won't help here (QEMU does not trap on unaligned accesses even when the physical CPU would). It is possible to make an x86 trap on unaligned accesses by setting bit 18 of the `eflags` register; this would look like this:

```
static inline void
set_align_check(int t)
{
        if (t) {
#if __i386__
                __asm__ __volatile__ (
                        "pushf\n\t"
                        "popl   %%eax\n\t"
                        "orl    $0x00040000, %%eax\n\t"
                        "pushl  %%eax\n\t"
                        "popf\n\t"
                        : : : "eax", "flags", "memory" );
#elif __x86_64__
                __asm__ __volatile__ (
                        "pushfq\n\t"
                        "popq   %%rax\n\t"
                        "orq    $0x0000000000040000, %%rax\n\t"
                        "pushq  %%rax\n\t"
                        "popfq\n\t"
                        : : : "rax", "flags", "memory" );
#else
#error Unsupported arch.
#endif
        } else {
#if __i386__
                __asm__ __volatile__ (
                        "pushf\n\t"
                        "popl   %%eax\n\t"
                        "andl   $0xFFFBFFFF, %%eax\n\t"
                        "pushl  %%eax\n\t"
                        "popf\n\t"
                        : : : "eax", "flags", "memory" );
#elif __x86_64__
                __asm__ __volatile__ (
                        "pushfq\n\t"
                        "popq   %%rax\n\t"
                        "andq   $0xFFFFFFFFFFFBFFFF, %%rax\n\t"
                        "pushq  %%rax\n\t"
                        "popfq\n\t"
                        : : : "rax", "flags", "memory" );
#else
#error Unsupported arch.
#endif
        }
}
```

I.e. some test framework could call `set_align_check(1)` before a function, and `set_align_check(0)` after that function has returned; if it did not crash with a `SIGBUS` then it did not do any unaligned access. This has some vexing characteristics, though:

  - On 32-bit x86, ABI only enforces 32-bit alignment, but `double` values are read as eight bytes and 4-byte alignment is not sufficient. Thus, totally standard and safe code may crash on this.
  - Library functions don't expect this flag and may crash (typically, `printf()` will crash). Most critical for crypto code is the low-level functions such as `memcpy()`, calls to which being possibly generated by the compiler without even the developer being aware of it.

It should be possible to write a test mechanism that uses this method, but it will require linking with special replacements for some library functions, possibly even linking with a special reduced libc replacement that offers _only_ a few functions (arguably, ability to run against a reduced libc can be viewed as a feature).

thomwiggers:
> If this warning is to be activated then I can probably patch all the code to use unions in declarations or something like that, but it is not going to be pretty.

Yeah, that's probably not something we want to put on both the people writing the implementations, and on the people 'consuming' them...

> QEMU won't help here (QEMU does not trap on unaligned accesses even when the physical CPU would).

Ok, so our testing infrastructure can't catch this – maybe we should invest more time in figuring out automated testing on actual hardware...

thomwiggers:
I don't think we can enable this then.

PR #182, opened 2019-06-03
Add Rainbow

mkannwischer:
Resolves #142 

#### Manually checked properties

* [ ] No stringification macros
* [ ] Output-parameter pointers in functions are on the left
* [ ] Negative return values on failure of API functions (within restrictions of FO transform).
* [ ] `const` arguments are labeled as `const`
* [ ] variable declarations at the beginning (except in `for (size_t i=...`)
* Optional:
  * [ ] All integer types are of fixed size, using `stdint.h` types (including `uint8_t` instead of `unsigned char`)
  * [ ] Integers used for indexing are of size `size_t`

PR #223, opened 2019-08-07
Enable valgrind exit errorcode and remove Rainbow

### LINKED ISSUE #220: Rainbow has undefined behaviour

thomwiggers:
It seems we were not doing anything with the valgrind results... D:

This PR reverts #182 because of #220. I think we should remove it for now, as it's not really suitable for any consumption at the moment.

Closes #220 

Reopen #142


thomwiggers:
This will definitely run into errors, as Rainbow has issues (#220)

mkannwischer:
LGTM. I'll look into the rainbow issue, but that will take a while.

PR #225, opened 2019-08-14
Adds AVX2 variants of Kyber512, Kyber768, Kyber1024

thomwiggers:
Adds the AVX2 variants of the Kyber packages.

This branch requires #224

#### Manually checked properties

* [x] No stringification macros
* [ ] Output-parameter pointers in functions are on the left
* [ ] Negative return values on failure of API functions (within restrictions of FO transform).
* [x] `const` arguments are labeled as `const`
* [x] variable declarations at the beginning (except in `for (size_t i=...`)
* Optional:
  * [ ] All integer types are of fixed size, using `stdint.h` types (
    * [x] including `uint8_t` instead of `unsigned char`
  * [x] Integers used for indexing are of size `size_t`

PR #224, opened 2019-08-13
Add support for specifying architecture and feature limits

thomwiggers:
This adds support for adding architecture-specific implementations to PQClean.

You would specify them as follows:

```yaml
    - name: avx2
      version: https://github.com/pq-crystals/kyber/commit/46e283ab575ec92dfe82fb12229ae2d9d6246682
      supported_platforms:
          - architecture: x86_64
            required_flags:
                - avx2
```

PR #226, opened 2019-08-21
Add 90s variants of Kyber

thomwiggers:
This adds the 90s variants of Kyber to PQClean.

This PR only includes the portable implementations, pending merge of #224 and #225 I will also add the AVX2 implementations.

I took the lazy way out and included the reference implementation's AES implementation. While this does result in some code duplication, I thought it easier than figuring out how the AES PRF works in Kyber.

* Optional:
  * [x] **Most** integer types are of fixed size, using `stdint.h` types (including `uint8_t` instead of `unsigned char`)
  * [x] **Most** Integers used for indexing are of size `size_t`


thomwiggers:
Ah I should still add duplicate consistency checks

thomwiggers:
I just remembered I haven't yet added duplicate consistency files....

PR #230, opened 2019-09-09
Fix and re-add Rainbow

### LINKED ISSUE #142: Re-Add Round 2 signature scheme Rainbow

mkannwischer:
Closes #142. 

Rainbow had some issues with undefined behaviour (#220) and was, thus, temporarily removed in #223. This was caused by casts from `uint8_t*` to `uint32_t*` in `blas_u32.c` which lead to crashes on some platforms. 
I've removed those casts and I've also refactored the finite field arithmetic quite a bit. Since it's now computing on nibbles (gf16) or bytes (gf256) instead of 8 or 4 field elements packed into a 32-bit integer, this is probably a lot slower than the previous implementation (which was also slow because I moved a lot of code from header files to source files which prevents inlining). 
However, this eliminates a lot of code which was quite hard to read (and not working on all platforms). 

For the actual changes, you might want to look at https://github.com/mkannwischer/rainbow-clean instead of the generated code here.


crypto_sign/rainbowIIIc-classic/clean/blas_u32.c
```diff
@@ -0,0 +1,87 @@
+#include "blas_u32.h"
+#include "gf.h"
+
+void PQCLEAN_RAINBOWIIICCLASSIC_CLEAN_gf256v_predicated_add_u32(uint8_t *accu_b, uint8_t predicate, const uint8_t *a, unsigned int _num_byte) {
+    uint32_t pr_u32 = ((uint32_t)0) - ((uint32_t)predicate);
+    uint8_t pr_u8 = pr_u32 & 0xff;
+
+    unsigned int n_u32 = _num_byte >> 2;
+    uint32_t *b_u32 = (uint32_t *)accu_b;
```
thomwiggers: This is unsafe behaviour if `accu_b` does not meet the `uint32_t` alignment requirements.

crypto_sign/rainbowIIIc-classic/clean/blas_u32.c
```diff
@@ -0,0 +1,87 @@
...
+    accu_b += (n_u32 << 2);
+    unsigned int rem = _num_byte & 3;
+    for (unsigned int i = 0; i < rem; i++) {
+        accu_b[i] ^= (a[i] & pr_u8);
+    }
+}
+
+void PQCLEAN_RAINBOWIIICCLASSIC_CLEAN_gf256v_add_u32(uint8_t *accu_b, const uint8_t *a, unsigned int _num_byte) {
+    unsigned int n_u32 = _num_byte >> 2;
+    uint32_t *b_u32 = (uint32_t *)accu_b;
```
thomwiggers: This is unsafe behaviour if `accu_b` does not meet the `uint32_t` alignment requirements.

mkannwischer:
> It seems there are still issues with pointer alignment remaining.

Yes, I'm working on that. It's harder than it seems to get rid of all those casts.

PR #235, opened 2019-09-19
Fixed sampler bug (update to new upstream Falcon code 2019-09-18).

pornin:
As announced on the pqc-forum mailing-list yesterday, there's a nasty bug (really, two bugs) in the sampler implementation in Falcon, leading to seemingly valid but vulnerable signatures (i.e. the signatures leak information on the private key). This PR includes the necessary fixes. Test vectors have (by necessity) changed, I updated the hashes in the META.yml files.

For details on the bugs, see https://eprint.iacr.org/2019/893 (section 6).

### DIFF
--- a/crypto_sign/falcon-1024/META.yml
+++ b/crypto_sign/falcon-1024/META.yml
@@ -20,4 +20,4 @@ auxiliary-submitters:
   - Zhenfei Zhang
 implementations:
     - name: clean
-      version: round two
+      version: 20190920
--- a/crypto_sign/falcon-1024/META.yml
+++ b/crypto_sign/falcon-1024/META.yml
@@ -4,8 +4,8 @@ claimed-nist-level: 5
 length-public-key: 1793
 length-secret-key: 2305
 length-signature: 1330
-nistkat-sha256: ad3d17869fdc05deae13ffa2ef26bde125b42f61b2dcd861a1ae20adcb2accc5
-testvectors-sha256: bd8076c13722d8c555c68fc6bd7763e1a9dd5483ee7c8d1c74dd2df459c72a40
+nistkat-sha256: ef2104e326221515621638ca03cd99802271bdd9907e2ae5fc7b8d19d696c584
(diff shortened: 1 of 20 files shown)


thomwiggers:
Thanks for the PR.

Ugh, `clang-tidy` can't make up it's mind on `ppc`, probably because it's a somewhat unstable version there (due to how the powerpc container is built from debian-ports).

As I don't really care much about that warning anyway, let's just disable it:

```patch
diff --git a/.clang-tidy b/.clang-tidy
index d141fd9..dafe2d1 100644
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,5 +1,5 @@
----
-Checks:          '*,-llvm-header-guard,-hicpp-*,-readability-function-size,-google-readability-todo,-readability-magic-numbers,-cppcoreguidelines-avoid-magic-numbers,-readability-isolate-declaration'
+--
+Checks:          '*,-llvm-header-guard,-hicpp-*,-readability-function-size,-google-readability-todo,-readability-magic-numbers,-cppcoreguidelines-avoid-magic-numbers,-readability-isolate-declaration,-readability-uppercase-literal-suffix'
 WarningsAsErrors: '*'
 HeaderFilterRegex: '.*'
 AnalyzeTemporaryDtors: false
```

thomwiggers:
Because the clang-tidy thing also affected #236, I've now created a separate PR that disables that warning: #237.

pornin:
Damn, it still fails. Apparently the fix on the 'u' / 'U' check has not come through yet...

... and it seems that that other fix has not been applied yet because of another "issue" (lack of an explicit cast from unsigned to int, required by clang-tidy on ppc only), which is fixed in this PR but was not in the other one. Or something like that.

thomwiggers:
https://github.com/PQClean/PQClean/pull/235 didn't get merged because it failed the same `bugprone-narrowing-conversion` check – which does seem like it's worthwhile to keep around. I've pushed a patch to #235, hopefully that'll silence clang-tidy (which is probably running LLVM8 due to the debian unstable/experimental nature of the ppc container – Debian no longer officially support ppc). After it gets merged, this PR needs to be rebased.

Meanwhile I should maybe look into getting all test containers to run the same version of LLVM/Clang/GCC.

PR #236, opened 2019-09-20
Document security issues

thomwiggers:
It seems wise to include _some_ form of guidance on what we know about
problems with certain (versions of) implementations.


mkannwischer:
LGTM

PR #239, opened 2019-10-12
qTESLA

sebastianv89:
Implementation of Round 2 qTESLA submission, both parameter sets p-I and p-III. Besides the usual cleaning up of the implementation, I spotted and fixed a bug (integer overflow). The qTESLA team has been informed as well.

PQClean#138

* Optional:
  * [x] Integers used for indexing are of size `size_t`


mkannwischer:
Seems like this is now failing due to a bug in my Kyber-90s code. 
I'm surprised that this was not caught by our tests earlier. 
I'll fix it so that you can re-base. Sorry about that.

mkannwischer:
Can you please rebase this on top of master?

sebastianv89:
Looks like the test only failed because some server was down during the test setup.

> distutils.errors.DistutilsError: https://files.pythonhosted.org/packages/1d/70/97966deebaeeda0b81d3cd63ba9f8ec929b838871ed17476de9d8159db3e/setuptools_scm-3.3.3-py2.py3-none-any.whl#sha256=1f11cb2eea431346d46589c2dafcafe2e7dc1c7b2c70bc4c3752d2048ad5c148 returned a bad status line. The server might be down, ''

thomwiggers:
Weird. I've restarted it, hopefully it'll pass now.

PR #240, opened 2019-10-15
Fix Clang 9.0.0 warnings

mkannwischer:
Clang 9.0.0 [1] and clang-tidy 9.0.0 [2] landed on my system recently and introduced a few warnings:

- [3] `security.insecureAPI.DeprecatedOrUnsafeBufferHandling` disallows the use of `memset` and `memcpy`. We use them a lot and I think in a "safe" way. I've disabled this warning. 
- [4] `bugprone-branch-clone` disallows having two equal branches in an if. We had that in Kyber and Rainbow - removed them as I think it's not necessary there. 
 
[1] https://releases.llvm.org/9.0.0/tools/clang/docs/ReleaseNotes.html
[2] https://releases.llvm.org/9.0.0/tools/clang/tools/extra/docs/ReleaseNotes.html
[3] https://clang.llvm.org/docs/analyzer/checkers.html#security-insecureapi-deprecatedorunsafebufferhandling-c 
[4] https://clang.llvm.org/extra/clang-tidy/checks/bugprone-branch-clone.html


mkannwischer:
I've rebased this on top of the falcon fix #242, I hope this passes now.

thomwiggers:
Weirdly, appveyor's kem tests appears to have also run LEDAKEM tests, which it shouldn't have. Rainbow being slow shouldn't be the problem, because Travis was able to do things in 25 minutes, similar for the CircleCI builds.

PR #246, opened 2019-10-22
Switch to gcc9.2 and xcode11.2 for MacOS builds

mkannwischer:
Our current builds on master, but also on #240  and #245  fail for GCC+MacOS because Travis has some issues with installing gcc 8. 
This works around by updating both gcc and xcode to the most recent version, which works fine. 
We want to upgrade eventually anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
   include:
     - name: "KEMs: MacOS + Clang"
       os: osx
-      osx_image: xcode10.2
+      osx_image: xcode11.2
       compiler: clang
       before_install:
         - pip3 install -r requirements.txt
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -47,14 +47,13 @@ matrix:
         - pip3 install -r requirements.txt
         - brew link gcc
         - export PATH="/usr/local/bin:$PATH"
-        - ln -s /usr/local/bin/gcc-8 /usr/local/bin/gcc
```
thomwiggers: This results in you running Clang instead of GCC

mkannwischer: good catch, thanks.

PR #247, opened 2019-10-24
Free SPHINCS+ hash state after use

0xa5a5:

* Optional:

### DIFF
--- a/crypto_sign/sphincs-haraka-128f-robust/clean/hash_haraka.c
+++ b/crypto_sign/sphincs-haraka-128f-robust/clean/hash_haraka.c
@@ -14,6 +14,11 @@ void PQCLEAN_SPHINCSHARAKA128FROBUST_CLEAN_initialize_hash_function(
     PQCLEAN_SPHINCSHARAKA128FROBUST_CLEAN_tweak_constants(hash_state_seeded, pub_seed, sk_seed, SPX_N);
 }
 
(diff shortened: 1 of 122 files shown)


0xa5a5:
As discussed there is no need for calling the destroy function in {NS}_thash, since the finalize function already does the hash context cleanup. Added a note to the {NS}_thash function.

crypto_sign/sphincs-haraka-128f-robust/clean/hash_haraka.c
```diff
@@ -14,6 +14,11 @@ void PQCLEAN_SPHINCSHARAKA128FROBUST_CLEAN_initialize_hash_function(
     PQCLEAN_SPHINCSHARAKA128FROBUST_CLEAN_tweak_constants(hash_state_seeded, pub_seed, sk_seed, SPX_N);
 }
 
+void PQCLEAN_SPHINCSHARAKA128FROBUST_CLEAN_destroy_hash_function(
+    hash_state *state) {
```
thomwiggers: This will need a comment to instruct the linter to ignore that `*state` could be const. See `initialize_hash_function` for SHAKE256.

thomwiggers:
Also, I should point out that the tests *completely skip* SPHINCS+, so you will need to show for yourself that these tests work by running the unittests.

I recommend running `pytest -nauto -k "sphincs"` on an appropriately fast machine ;-)

PR #248, opened 2019-10-24
Add {Baby,Mama,Papa}Bear reference implementations

### LINKED ISSUE #127: Add Round 2 KEM ThreeBears

leonbotros:
This PR adds the reference implementation of ThreeBears by Mike Hamburg.
The code that generates these implementations is adjusted from [rainbow-clean](https://github.com/mkannwischer/rainbow-clean) and available at: [threebears-clean](https://github.com/leonbotros/threebears-clean). 
The implementations match the newest version (see, https://sourceforge.net/projects/threebears/), 
that includes implicit rejection. Feedback is welcome!

Closes #127. 

* [X] No stringification macros
* [X] Output-parameter pointers in functions are on the left
* [X] `const` arguments are labeled as `const`
* [X] variable declarations at the beginning (except in `for (size_t i=...`)
* Optional:
  * [X] Integers used for indexing are of size `size_t`

### DIFF
--- a/crypto_kem/babybear/clean/Makefile
+++ b/crypto_kem/babybear/clean/Makefile
@@ -6,7 +6,7 @@ HEADERS = api.h melas_fec.h params.h ring.h threebears.h
 OBJECTS = kem.o melas_fec.o ring.o threebears.o 
 
 
-CFLAGS=-O3 -Wcast-align=strict -Wall -Wconversion -Wextra -Werror -Wpedantic -Wvla -Wmissing-prototypes -Wredundant-decls -std=c99 -I../../../common $(EXTRAFLAGS)
+CFLAGS=-O3 -Wall -Wextra -Wpedantic -Werror -Wmissing-prototypes -Wredundant-decls -std=c99 -I../../../common $(EXTRAFLAGS)
 
 all: $(LIB)
 
--- a/crypto_kem/babybear/clean/Makefile
+++ b/crypto_kem/babybear/clean/Makefile
@@ -6,7 +6,7 @@ HEADERS = api.h melas_fec.h params.h ring.h threebears.h
 OBJECTS = kem.o melas_fec.o ring.o threebears.o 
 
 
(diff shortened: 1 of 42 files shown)


crypto_kem/babybear/clean/ring.h
```diff
@@ -0,0 +1,31 @@
...
+/* Multiply and accumulate c = c + a*b */
+void PQCLEAN_BABYBEAR_CLEAN_mac(gf_t c, const gf_t a, const gf_t b);
+
+/* Reduce ring element to canonical form */
+void PQCLEAN_BABYBEAR_CLEAN_canon(gf_t c);
+
+/** Return the i'th limb of the modulus */
+static inline limb_t modulus(size_t i) {
+    return (i == DIGITS / 2) ? LMASK - 1 : LMASK;
+}
```
mkannwischer: This should be moved out of the header file.

crypto_kem/babybear/clean/Makefile
```diff
@@ -0,0 +1,21 @@
+# This Makefile can be used with GNU Make or BSD Make
+
+LIB=libbabybear_clean.a
+
+HEADERS = api.h melas_fec.h params.h ring.h threebears.h
+OBJECTS = kem.o melas_fec.o ring.o threebears.o 
+
+
+CFLAGS=-O3 -Wcast-align=strict -Wall -Wconversion -Wextra -Wpedantic -Wvla -Wmissing-prototypes -Wredundant-decls -std=c99 -I../../../common $(EXTRAFLAGS)
```
mkannwischer: @thomwiggers, does this mean that -Wcast-align=strict is now supported by all of our targets?
If yes, then we should add this to all the implementations.

thomwiggers: I don't think it is; there's still false positives probably.

leonbotros: I think `-Werror` is missing here

mkannwischer:
LGTM, thanks @leonbotros!

thomwiggers:
Travis-CI's arm64 builds appear to be working again, let's hope they get this show on the road.

PR #250, opened 2019-11-13
Rainbow: memcpy -> memmove since buffers may overlap

mkannwischer:
Our tests on master failed (among other reasons) because rainbow had some overlapping buffers passed to `memcpy`: https://circleci.com/gh/PQClean/PQClean/8556 

It's right. This should be a `memmove`.


thomwiggers:
This fix has been verified on my computer.

PR #253, opened 2019-11-28
SPHINCS+ optimized implementations

### LINKED ISSUE #249: Improve documentation for hashing API in header file sha2.h
_Originally posted by @thomwiggers in https://github.com/PQClean/PQClean/pull/247#issuecomment-545841257_

thomwiggers:
Include SPHINCS+ optimized implementations: AVX2 and AESNI.

Also updates the `clean` versions, because they're being generated from a shared codebase.

Supersedes #247: closes #247 when merged. Also fixes #249 

* Optional:


thomwiggers:
Shit, this should've been a draft PR.

thomwiggers:
* [ ] Tested on Windows
  * [ ] 64-bit
  * [ ] 32-bit
* [ ] Tested on Linux:
  * [x] amd64
    * [ ] Clang
    * [ ] GCC
  * [ ] i386
    * [ ] Clang
    * [ ] GCC
  * [ ] arm32
    * [ ] Clang
    * [ ] GCC
  * [ ] arm64
    * [ ] Clang
    * [ ] GCC
  * [ ] ppc
    * [ ] Clang
    * [ ] GCC

thomwiggers:
See also https://github.com/thomwiggers/sphincsplus/tree/pqcleanup

crypto_sign/sphincs-haraka-128f-robust/clean/address.h
```diff
@@ -1,13 +1,13 @@
...
+#define PQCLEAN_SPHINCSHARAKA128FROBUST_CLEAN_ADDRESS_H
 
 #include <stdint.h>
 
-#define SPX_ADDR_TYPE_WOTS 0
-#define SPX_ADDR_TYPE_WOTSPK 1
-#define SPX_ADDR_TYPE_HASHTREE 2
-#define SPX_ADDR_TYPE_FORSTREE 3
-#define SPX_ADDR_TYPE_FORSPK 4
+#define PQCLEAN_SPHINCSHARAKA128FROBUST_CLEAN_ADDR_TYPE_WOTS 0
```
mkannwischer: Why namespace these constants?

thomwiggers: They always had an SPX_ prefix; which was also the prefix I chose to add to all symbols. They simply got namespaced with all the rest.

mkannwischer: ok, I don't think this is necessary or consistently done in the other implementations in PQClean, but it doesn't hurt.

PR #265, opened 2020-02-05
Move keys in crypto_sign/functest.c to the heap

thomwiggers:
Having the keys on the stack increases the stack space consumption by
quite a bit, and this in turn results in the sanitizer tests failing
for Rainbow.

Moving the keys to the heap in the test seems like a harmless change.


thomwiggers:
See also https://app.circleci.com/jobs/github/PQClean/PQClean/9632

mkannwischer:
Would it also make sense to do this for KEMs? 
I wouldn't want the functest.c files to diverge too much from each other.

PR #266, opened 2020-02-06
Put all common primitives on the heap

### LINKED ISSUE #264: Abstract away from primitive states being on the stack
The current API for AES and FIPS202 still completely relies on the full state being on the stack. This is not a very friendly assumption when you want to allow swapping out the implementation for these primitives, and may be something we want to fix, but it's there for now.

Solving this would require adding `_copy` and `_free` functions to all primitives exposed by the PQClean common API. We should then also somehow make sure that these `_copy` and `_free` functions are used... This might already require that we put the `common/` implementations on the heap to allow Valgrind to find the leaks.

See also #207, #247, https://github.com/open-quantum-safe/liboqs/pull/600 for examples where this came up already.

thomwiggers:
This helps enforce proper use of their API.

Fixes #264

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,7 +57,12 @@ See the section [API](#API) below.
     astyle --project crypto_kem/yourschemename/clean/*.[ch]
     ```
     4. You may run the tests in the `tests/` folder. See the `README` for how to run the test suite.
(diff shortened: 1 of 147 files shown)


thomwiggers:
Blocked by #267

thomwiggers:
It's amazing how much difference a fresh perspective can make in finding the bugs that are haunting you.

PR #270, opened 2020-02-15
Fix NewHope CCA verify function

### LINKED ISSUE #269: Bug in NewHope CCA verify function
https://github.com/mupq/pqm4/issues/132 reports a bug in the NewHope verify function that was introduced by me by adding a wrong cast when integrating it into PQClean. 
Sorry about that. I'll fix this asap.

mkannwischer:
Fixes #269.
https://github.com/mupq/pqm4/issues/132 reported that the NewHope verify function does not actually return 0 or 1, but 0 or -1, which consequently breaks the cmov in the FO transform (in case of a re-encryption failure). 
This bug was introduced when I integrated this into PQClean. I added a wrong cast. Without the cast, it works correctly on my machine. 
If I recall correctly, this failed on some of the other platforms. Let's see.

### DIFF
--- a/crypto_kem/newhope1024cca/clean/reduce.c
+++ b/crypto_kem/newhope1024cca/clean/reduce.c
@@ -5,7 +5,7 @@ static const uint32_t qinv = 12287; // -inverse_mod(p,2^18)
 static const uint32_t rlog = 18;
 
 /*************************************************
-* Name:        verify
+* Name:        montgomery_reduce
 *
 * Description: Montgomery reduction; given a 32-bit integer a, computes
 *              16-bit integer congruent to a * R^-1 mod q,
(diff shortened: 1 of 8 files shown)


mkannwischer:
Somehow all the tests got skipped in our Windows builds. Any ideas @thomwiggers?

thomwiggers:
That's strange indeed, no idea...

thomwiggers:
I'm logged in via RDP on one of the Appveyor boxes, trying to run `permit_test`, but that says that it should be working...

thomwiggers:
the problem is that the paths returned by `git` use forward slashes, and the paths on Windows are backslashes... so the `PQCLEAN_ONLY_DIFF` helper things it's not one of the changed paths.

thomwiggers:
```
verify.c(25): error C2220: the following warning is treated as an error
verify.c(25): warning C4146: unary minus operator applied to unsigned type, result still unsigned
```

thomwiggers:
If you rebase on master to include #277 it'll pass

thomwiggers:
Rebased and force-pushed

PR #272, opened 2020-02-16
Fix memory leak in Kyber

dstebila:
Noticed by https://github.com/open-quantum-safe/liboqs/pull/600, but this PR covers more cases than that PR.


thomwiggers:
Strange that this was not caught

PR #274, opened 2020-02-19
Fix detecting of to-be-tested changes on Windows

thomwiggers:
The `git diff --names-only` command returned `paths/like/this` also on Windows. This broke the detecting of the common prefix to determine if tests needed to be run.

See also #270


thomwiggers:
I'm force-merging this, as most of the tests already passed and testing on windows is pointless without it.

PR #279, opened 2020-03-13
Split aes*_keyexp up into ecb and ctr variants

dstebila:

* Optional:


dstebila:
Is this just a random CI failure or something else? It's in something that I didn't touch at all.

dstebila:
@thomwiggers I'm not sure what to do about the failure in https://travis-ci.com/github/PQClean/PQClean/jobs/305201629, nor whether the build errors / timeouts are problematic or not.

thomwiggers:
Pretty sure the `va_list` stuff is buggy behaviour of clang-tidy... Re-running tends to help.

thomwiggers:
I've added `PQCLEAN_SKIP_TESTS` to most of the emulated builds, for tests that should be the same across platforms anyways. Hopefully that saves a lot of build time.

mkannwischer:
What was the rationale for this change?

thomwiggers:
`_keyexp` is also more or less the `_new` or `_init` for the AES state struct. Depending on what the underlying AES implementation is, this may (our stack based API) or may not (OpenSSL) be the same operation.

PR #288, opened 2020-04-15
Add MQDSS AVX2 implementations

thomwiggers:
Add MQDSS AVX2 implementations

It turned out that they're quite straightforward adaptions using only the intrinsics and not any complicated assembly files.

Only files needing review (rest was copied from the `clean` version):

* gf31.c
* gf31.h (same between both MQDSS-48 and MQDSS-64)
* mq.c (different between both versions)
* Makefiles

* Optional:

### DIFF
--- a/crypto_sign/mqdss-48/META.yml
+++ b/crypto_sign/mqdss-48/META.yml
@@ -16,3 +16,12 @@ auxiliary-submitters:
 implementations:
     - name: clean
       version: https://github.com/joostrijneveld/MQDSS/commit/00608d7610262ff07b1834885d32bc3fd27ef5e1
+    - name: avx2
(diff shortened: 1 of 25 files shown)


thomwiggers:
The appveyor failure seems inconsequential.

mkannwischer:
Are these failing duplicate consistency checks lying?

thomwiggers:
It was not.

thomwiggers:
Locally I had some issues with `clang-tidy` complaining about variables not being initialised at declaration, but also in the original code... This might just be a version thing though.

thomwiggers:
I rebased to make it hopefully pass CI this time (with the new speed improvements); you can click the green button if you want.

PR #290, opened 2020-04-21
Update required_flags for Dilithium's META.ymls

xvzcf:
Updated `required_flags` in Dilithium META.yml files to match the flags used in corresponding Makefiles.

### DIFF
--- a/crypto_sign/dilithium2/META.yml
+++ b/crypto_sign/dilithium2/META.yml
@@ -27,4 +27,5 @@ implementations:
                 - Linux
             required_flags:
                 - avx2
-                - bmi2
+                - bmi1
+                - popcnt
--- a/crypto_sign/dilithium3/META.yml
+++ b/crypto_sign/dilithium3/META.yml
@@ -27,4 +27,5 @@ implementations:
                 - Linux
             required_flags:
                 - avx2
-                - bmi2
+                - bmi1
+                - popcnt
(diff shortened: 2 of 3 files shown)


crypto_kem/mceliece348864/META.yml
```diff
@@ -34,7 +34,7 @@ implementations:
             - Linux
             - Darwin
           required_flags:
-            - sse4_1
+            - sse4.1
```
thomwiggers, line 37: Ah, this is actually necessary due to it being reported as `sse4_1` by `/proc/cpuinfo`...

crypto_kem/mceliece348864f/META.yml
```diff
@@ -34,8 +34,8 @@ implementations:
             - Linux
             - Darwin
           required_flags:
-            - sse4_1
-            - bmi1
+            - sse4.1
+            - bmi
```
thomwiggers, line 38: Same story here; it's `bmi1` in cpuinfo :(

crypto_sign/dilithium2/META.yml
```diff
@@ -27,4 +27,5 @@ implementations:
                 - Linux
             required_flags:
                 - avx2
-                - bmi2
+                - bmi
```
thomwiggers, line 30: this should be `bmi1` then; `bmi2` is indeed incorrect.

thomwiggers:
Thanks, much appreciated!

PR #300, opened 2020-05-29
Use Github Actions to run tests (+ many SPHINCS+ AVX2 fixes for Windows)

thomwiggers:
This PR tries to use Github actions to run the tests. Hopefully it will make running the tests a bit more predicable. I've split the workflows per scheme (there's a template and generator included) because that allows me to use Github's logic to only run the tests for those codepaths that have been changed.

Github also has more modern chips that the tests are running on, as they use Azure VMs for their Windows / Linux CI boxes (at least Haswell). OS X boxes are hosted by MacStadium, and have a little less documentation, but they seem more modern than the ones provided by Travis [e.g. see this run](https://github.com/thomwiggers/test-macosrunner/runs/757289189?check_suite_focus=true) (seems at least Ivy Bridge)

* [x] Linux based tests
  * [x] AMD64
  * [x] 32-bit intel
  * [x] Emulated armhf
  * ~~ARM64~~ Not supported by Github's runners, but self-hosted runners support this!
* [x] MacOS based tests
   * [x] GNU GCC
   * [x] Apple LLVM/Clang
* [x] Windows based tests
   * [x] 32-bit
   * [x] 64-bit


thomwiggers:
Now also runs tests on MacOS and Windows.

thomwiggers:
Tests have only been generated for Kyber512, would of course need to generate them for all the schemes.

thomwiggers:
Suggested filter for reviewing https://github.com/PQClean/PQClean/pull/300/files?file-filters%5B%5D=.j2&file-filters%5B%5D=.md&file-filters%5B%5D=.py

.github/workflows/generate_workflows.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
+
+from pathlib import Path
+
+import jinja2
+
+
```
thomwiggers, line 7: Should add a test that checks if the workflows are up-to-date

thomwiggers:
> API requests - You can execute up to 1000 API requests in an hour across all actions within a repository. If exceeded, additional API calls will fail, which might cause jobs to fail. This limit also applies to self-hosted runners.

Oops.

mkannwischer (dismissed):

thomwiggers:
Somehow the `test_makefile_dependencies` test is failing for SPHINCS on windows but I can't reproduce it...

thomwiggers:
The Python setup is failing because of https://github.com/actions/setup-python/pull/144, this needs to be included in a release before we can continue.

thomwiggers:
It took a while but I finally tracked down why the SPHINCS+ shake256 AVX2 builds kept failing: it got into a fight over the need to update the dependencies in the `test_makefile_dependencies` test.

thomwiggers:
Everything passed!

PR #305, opened 2020-07-23
Remove Non-Round3 Schemes

mkannwischer:
NIST announced the Round 3 finalists and alternate candidates today:
https://groups.google.com/a/list.nist.gov/d/msg/pqc-forum/0ieuPB-b8eg/Cl7Ji8TpCwAJ

Shall we remove the schemes from PQClean that did not make it to Round 3 (either as finalists or alternate candidates)? 
That include threebears, ledakem, newhope, mqdss, and qtesla. 

Do you see any reason why we should keep them?


thomwiggers:
While Round 3 is a thing now, there may still be things going on that consume implementations from PQClean that use Round-2 candidates. However, I suppose those people can also still check out older versions of the repository to get those implementations.

I'm ambivalent.

cryptojedi:
Thom Wiggers <notifications@github.com> wrote:


> While Round 3 is a thing now, there may still be things going on that
> consume implementations from PQClean that use Round-2 candidates.
> However, I suppose those people can also still check out older
> versions of the repository to get those implementations.
> 
> I'm ambivalent.

I would remove all non-round-3 schemes from PQClean, but document what
the latest version (commit) is containing those schemes.

thomwiggers:
The right feature to do that is a git tag :wink:

Issue #251, opened 2019-11-13
Kyber-AVX2 doesn't work on OS X, Windows


thomwiggers:
The Kyber ASM code doesn't work on OS X (macros don't seem to be supported) nor on Windows (Microsoft compiler has other syntax).

thomwiggers:
The MacOS tests have not been running because the Travis Mac cpus are quite old and don't have AVX2.

thomwiggers:
This can probably be solved in the same way as Dilithium. The actual problem is that C symbols are prefixed with an underscore on MacOS

thomwiggers:
OS X compatibility changes required:

* Add underscore equivalents of symbols in assembly functions or do it dynamically with a macro (the C ABI starts with an underscore on OS X) https://github.com/PQClean/PQClean/blob/master/crypto_sign/dilithium2/avx2/ntt.S#L44-L45* Make sure any symbols referred to in ASM code that are defined in C files (constants?) are wrapped in a macro like https://github.com/PQClean/PQClean/blob/master/crypto_sign/dilithium2/avx2/ntt.S#L47-L49

mouse07410:
Since this issue appears to be quite old - any chance to get the fix sooner...?

```
$ make SCHEME=Kyber1024 IMPLEMENTATION=avx2
cd ../crypto_kem/Kyber1024/avx2 && /Applications/Xcode.app/Contents/Developer/usr/bin/make
clang -mavx2 -mbmi2 -mpopcnt -O3 -Wall -Wextra -Wpedantic -Werror -Wmissing-prototypes -Wredundant-decls -std=c99 -I../../../common  -c -o kem.o kem.c
clang -mavx2 -mbmi2 -mpopcnt -O3 -Wall -Wextra -Wpedantic -Werror -Wmissing-prototypes -Wredundant-decls -std=c99 -I../../../common  -c -o poly.o poly.c
clang -mavx2 -mbmi2 -mpopcnt -O3 -Wall -Wextra -Wpedantic -Werror -Wmissing-prototypes -Wredundant-decls -std=c99 -I../../../common  -c -o polyvec.o polyvec.c
yasm -m amd64 -f macho64 -p gas  -o fq.o fq.s
yasm -m amd64 -f macho64 -p gas  -o shuffle.o shuffle.s
[...]

PR #307, opened 2020-07-31
Update Kyber from upstream

thomwiggers:
Makes Kyber-AVX run on MacOS (#251)

* Optional:


mouse07410:
I confirm that with this PR merged, Kyber AVX2 builds successfully on macOS Catalina with the latest released Xcode.

thomwiggers:
Your patch would break Linux support, as the symbols are exported from C without prefix `_` there.

The other test failing is due to #325 landing, and I will get around to rebasing and fixing that.

mouse07410:
OK, **now** it passes the tests. ;-)

Thank you!

thomwiggers:
@mkannwischer although round-3 kyber will soon be a thing, I think we should still merge this as it will probably take a while before we get Kyber-R3 in a mergable shape.

crypto_kem/kyber512/avx2/cdecl.inc
```diff
@@ -0,0 +1,30 @@
+#ifndef PQCLEAN_DILITHIUM2_AVX2_CDECL
```
jschanck, line 1: Bad namespacing here and in other cdecl.inc files.

thomwiggers:
Replaced by #349

Issue #309, opened 2020-08-07
Undefined behaviour in FireSaber


claucece:
Dear, PQClean project,

Thanks so much for this project!

While running the FiraSaber KEM with liboqs (which copies it from this library: https://github.com/open-quantum-safe/liboqs/blob/master/scripts/copy_from_pqclean/copy_from_pqclean.yml#L173, https://github.com/open-quantum-safe/liboqs/blob/master/src/kem/saber/CMakeLists.txt#L3) with the `-fsanitize=undefined` flag, this undefined behaviour was flagged (coping the output of the liboqs library):

```
tests/test_kem FireSaber-KEM    
Configuration info
==================
Target platform:  x86_64-Darwin-18.7.0
Compiler:         clang (11.0.0 (clang-1100.0.33.17))
OQS version:      0.4.0-rc1
Git commit:       477ef3f60cd04f9b72b499d000a8f77ae99c0d89 (+ local modifications)
OpenSSL enabled:  Yes (OpenSSL 1.1.1g  21 Apr 2020)
AES:              OpenSSL
SHA-2:            OpenSSL
SHA-3:            OpenSSL

================================================================================
Sample computation for KEM FireSaber-KEM
================================================================================
../src/kem/saber/pqclean_firesaber_clean/poly_mul.c:50:44: runtime error: signed integer overflow: 57354 * 40101 cannot be represented in type 'int'
shared secrets are equal
```

which corresponds to line `https://github.com/PQClean/PQClean/blob/master/crypto_kem/firesaber/clean/poly_mul.c#L50` over here.

I'm using clang 11.0.0 on x86_64-Darwin-18.7.0.

thomwiggers:
It needs a few casts, Kyber had the same problem but running into it is a matter of probability. The Kyber code should have a fix in the same file and function. 

PRs are more than welcome :)

(Greetings from [Wanneperveen](https://en.visitweerribbenwieden.com/))

claucece:
> PRs are more than welcome :)

I'll send the PR over the weekend ;)

> (Greetings from Wanneperveen)

Oh, so jealous! Enjoy! ;)

thomwiggers:
Closed by #310

PR #310, opened 2020-08-11
Fix overflow in multiplication in Saber

claucece:
Fix the undefined behaviour on Saber as defined on #309. Based on: https://github.com/PQClean/PQClean/commit/39fa1ccfc7ceb38cdb0777eed6b3562c8c7eeaa6#diff-4db6c2719416e99dc8aa9cb30206d852

@thomwiggers , what do you think?


thomwiggers:
Not sure if this gets all of them, but it will help

PR #312, opened 2020-08-15
Add AVX2 implementations of NTRU

jschanck:
Adds the AVX2 implementations of ntruhps{2048509, 2048677, 4096821} and ntruhrss701 which are in SUPERCOP.

The ntruhps* implementations (and, eventually, NTRU Prime) would benefit from an AVX2 implementation of crypto_sort, but this probably shouldn't be duplicated across implementations. I'll look into putting some crypto_sort implementations into PQClean/common.


jschanck:
The AVX2 implementation of NTRU declares its polynomial type with `__attribute__((aligned(32)))` and uses `vmovdqa` instructions. This is failing in test_char.py because pycparser chokes on the `__attribute__`.

One solution: define `__attribute__(x)` away in test_char.py, e.g.
`            cpp_args=[
                '-D__attribute__(x)=',
                '-E',
...
            ]
`

That said, I don't see any other uses of `__attribute__` in the code. So maybe there is an alternative way to enforce alignment?

thomwiggers:
`__attribute__` is not portable. The "correct" way to enfore alignment is to create a union with a correctly aligned type. This can have the upside that your casts become more clear, but it definitely is a bit of work.

For an example, see https://github.com/PQClean/PQClean/blob/4ea4b478baa436465fbf890e385f46d432d9a337/crypto_sign/dilithium2/avx2/alignment.h

jschanck:
Thanks, Thom. That did the trick.

dstebila:
Need to delete ntruhps2048509/avx2/.Makefile.swp.

dstebila:
Are any of the files the same, or meant to be the same (modulo namespacing), as files in the clean versions, or across the avx2 implementations of the different variants?  We have the [duplicate consistency](https://github.com/PQClean/PQClean/tree/master/test/duplicate_consistency) mechanism which can be used to track that and diffs the intended-same files to ease code review.

jschanck:
There were some upstream changes to the NTRU code. No point in reviewing this PR, so I'm going to close and resubmit.

PR #313, opened 2020-08-19
Add common/crypto_sort

jschanck:
This PR makes djbsort available in the common directory. The constant time sorting routines from this library are used in the implementations of McEliece, NTRU, and NTRU Prime.

PQClean currently duplicates code across the three schemes. This doesn't create much extra work at the moment, but the sorting routines are a significant optimization target and we can expect some churn in the implementations (especially as PQClean gets more platform-specific implementations).

SUPERCOP factors out crypto_sort in this way as well.

### DIFF
--- a/common/crypto_sort/int32/portable3/crypto_sort.c
+++ b/common/crypto_sort/int32/portable3/crypto_sort.c
@@ -1,0 +1,46 @@
+#include "crypto_sort_int32.h"
+#define int32 int32_t
+#define int32_sort crypto_sort_int32
+
+#define int32_MINMAX(a,b) \
+    do { \
+        int32 ab = b ^ a; \
+        int32 c = b - a; \
+        c ^= ab & (c ^ b); \
+        c >>= 31; \
+        c &= ab; \
+        a ^= c; \
+        b ^= c; \
+    } while(0)
+
+void int32_sort(int32 *x, size_t n) {
+    size_t top, p, q, r, i;
+
+    if (n < 2) {
+        return;
+    }
+    top = 1;
+    while (top < n - top) {
+        top += top;
+    }
+
+    for (p = top; p > 0; p >>= 1) {
+        for (i = 0; i < n - p; ++i) {
+            if (!(i & p)) {
+                int32_MINMAX(x[i], x[i + p]);
+            }
+        }
+        i = 0;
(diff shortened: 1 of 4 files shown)


common/crypto_sort/int32/portable3/crypto_sort.c
```diff
@@ -0,0 +1,46 @@
+#include "crypto_sort_int32.h"
```
thomwiggers, line 1: Why are there two `portable` implementations?

common/crypto_sort_int32.h
```diff
@@ -0,0 +1,16 @@
+#ifndef CRYPTO_SORT_INT32_H
+#define CRYPTO_SORT_INT32_H
+
+#include <stdint.h>
+
+#ifdef __cplusplus
```
thomwiggers, line 6: This should probably be removed, we don't support C++ and I'm not sure how our linters would interact with this header file

thomwiggers (commented):
I'm not sure how I feel about this.

The `common` library/folder is supposed to provide functions to facilitate testing, not to provide the fastest-speed implementations of primitives. Having AVX2 versions in `common` seems a bit contrary to this. Any implementation relying on `common` will be slowed down significantly, because we use the heap extensively there even though it's not necessary, sometimes this leads to `malloc` in very tight loops in implementations. However, that helps us with leak checking. 

The goal for the `clean` implementations (because of AVX2's inherent complexity we allow a little bit more there) is to be able to compile them with ``gcc -I common common/*.c crypto_TYPE/SCHEME/clean/*.c``. This helps people get started quickly. One of the things I dislike about SUPERCOP is all the weird corners unexpected pieces of code seem to be coming from — where `crypto_sort` was coming from was a big source of confusion when I extracted Classic McEliece from SUPERCOP.

On the other hand, you do have a point that, exactly because sorting can be viewed as a performance-critical "primitive", anyone adopting code from PQClean would possibly be well-suited to select code that works best for their target platform, rather than just whatever (portable) code happened to be integrated into the scheme folder in PQClean. As I mentioned in #308, `common` is fairly in the "bring your own code" department.

jschanck:
OK, I'm satisfied with your reason and I'm closing this PR. Thanks.

PR #314, opened 2020-08-24
Update NTRU and add AVX2 implementations

jschanck:
version: https://github.com/jschanck/ntru/tree/ba431bb8

Built using the new "package_pqclean.sh" script.


thomwiggers:
Did you test this on Windows yourself? Otherwise I should do some manual testing on Windows boxes because I think the CI service doesn't have AVX2 unfortunately :(

thomwiggers:
Nevermind the AVX2 impl doesn't support Windows anyway.

Same still goes for MacOS, because Travis doesn't support AVX2 either, but I don't have access to a Mac we'll probably do that some other time.

crypto_kem/ntruhps4096821/clean/sample.c
```diff
@@ -1,27 +1,30 @@
...
 }
 
 void PQCLEAN_NTRUHPS4096821_CLEAN_sample_rm(poly *r, poly *m, const unsigned char uniformbytes[NTRU_SAMPLE_RM_BYTES]) {
+
     PQCLEAN_NTRUHPS4096821_CLEAN_sample_iid(r, uniformbytes);
     PQCLEAN_NTRUHPS4096821_CLEAN_sample_fixed_type(m, uniformbytes + NTRU_SAMPLE_IID_BYTES);
 }
 
+
+#include "crypto_sort_int32.h"
```
thomwiggers, line 16: Should this not be at the top of this file? (I think one of the linters is supposed to not like this...)

thomwiggers (changes requested):
One teeny tiny nitpick, otherwise it looks great.

jschanck:
Thanks for the review!

Upstream has been tested on macOS, but I've only personally run the PQClean version on Linux.

thomwiggers (approved):

PR #316, opened 2020-08-27
ntruhrss701/avx2: fix non-PIC call

jschanck:
Fixes an issue with https://github.com/open-quantum-safe/liboqs/pull/817


thomwiggers (approved):
Not sure when exactly calling with `@plt` is or isn't necessary. Should we make sure to run our tests with `-fPIC`, `-fPIE`?

jschanck:
The issue (as far as I understand it) is linking with -shared. Testing with -fPIC or -fPIE would not be enough.

PR #317, opened 2020-08-29
Add sntrup{653,761,857} and ntrulpr{653,761,857}

jschanck:
Exported from SUPERCOP-20200826 using the scripts at:
https://github.com/jschanck/pqclean-package-ntruprime

* Optional:


jschanck:
FYI this is only failing (with gcc/clang) in test_nistkat. I don't know why. Could be that the code from supercop isn't generating the same KATs as the version submitted to NIST for round 2. Could be a change I introduced. The clean/ and avx2/ versions generate the same KAT values, so if it's a change I introduced it's probably in crypto_sort or crypto_stream. I'll look into it on Monday.

jschanck:
Dan informed me that the build was failing test_nistkat because of a quirk in the NIST randombytes API. See section "Randomness and known-answer tests" of https://blog.cr.yp.to/20170719-pqbench.html. In short, one call to randombytes(..., k*L) is not the same as k calls to randombytes(..., L). Dan views this as a bug, and I'm inclined to agree. 

My patch here is to NTRU Prime rather than randombytes. There is an associated performance penalty, which seems unfair. so I'll look into other solutions. Maybe a shim in the test instead. Would we be open to that?

dstebila:
Not sure what the right answer is for PQClean.  In liboqs, we have a run-time swappable RNG, one of which is `OQS_randombytes_nist_kat`. We could create an alternate `OQS_randombytes_nist_kat_djb` and then hardcode a special case in the KAT harness to use the `djb` variant for NTRU Prime.

jschanck:
Current passing build is generated from https://github.com/jschanck/pqclean-package-ntruprime/commit/f3670c4bdb822df456ede3cfcf11c8878abef9b3

Regarding test_nistkat: One solution is to use the "alternative" KAT values. Dan has a good point regarding the NIST randombytes API, and he's illustrated that point by making a slightly unusual pattern of randombytes calls in the NTRU Prime reference implementation. If we assume that NIST has no interest in fixing their API, then it makes sense for the normative KATs to be those that come from the pattern of calls to randombytes that implementations will actually use (a single call for all of the random bytes).

If we want to go that route the generation scripts are here:
https://github.com/jschanck/pqclean-package-ntruprime/commit/d3a087f8424b065f2fa3afbe9448862645b1b59e

thomwiggers:
Hm, I'm having a little bit of trouble parsing these patch files by eye, but;

```
 randombytes((unsigned char *) L,sizeof L);
```

`L` seems a `uint32_t` size type here. This means that this might be prone to endianness problems when interpreting as `uint32_t` again on `ppc`.

jschanck:
Should be OK. It does 
```
randombytes((unsigned char *) L,sizeof L);
crypto_decode_pxint32(L,(unsigned char *) L);
```

The crypto_decode_pxint32 procedure is endianness-independent.

thomwiggers:
> then it makes sense for the normative KATs to be those that come from the pattern of calls to randombytes that implementations will actually use (a single call for all of the random bytes).

I think that is the best option; if I understand what you're saying correctly then otherwise (without fixing the PRNG) you'd force alternative implementations to use the same pattern of calls.

Fixing the `randombytes` for the `test_nistkat` is tempting, but I think it might lead to confusing for other consumers that might integrate things and then test it against the "reference" test framework (like OQS), without the patches.

jschanck:
OK, sounds like we're in agreement that I should add the changes from my altkats branch to this PR.

Do we want to wait for other opinions?

To recap:
- The NIST submission and supercop implementations of NTRU Prime are not NIST-KAT compatible.
(The NIST submission makes p calls to randombytes(..., 4), and supercop makes one call to randombytes(..., 4*p).)
- This PR is based on the supercop code, but includes a patch for NIST-KAT compatibility.
- My altkats branch breaks NIST-KAT compatibility. It passes test_nistkat because the hashes in META.yml come from the altkats code itself (i.e. the test is meaningless).
- It's OK that the test is meaningless, because this PR currently passes test_nistkat with the true hashes, and the diff to altkats is small enough to check by hand.

thomwiggers:
> Do we want to wait for other opinions?

I think we can just proceed. By the time you've pushed those changes, CI passed, and we've found time to press the merge button, there's been ample time for people to respond.

jschanck:
Failing because of a timeout. Is there a way to give this more time to run?

thomwiggers:
I don't think those tests should hang, possibly something got stuck in an infinite loop somewhere. I've restarted the workflow.

I really want to move most CI over to GH with #300, that'd allow us to run the tests per scheme in a much more reliable way.

thomwiggers (approved):

thomwiggers:
Much appreciated!

PR #324, opened 2020-09-07
New HQC and HQC-RMRS from upstream

### LINKED ISSUE #291: Update HQC
>  - We improved the DFR analysis of our protocol: as a consequence, we managed to decrease the size of the public key by 3%
>
> - We introduced an alternative and more efficient encoding and decoding algorithm based on concatenating Reed-Muller and Reed-Solomon codes. This new decoding variant does not change the protocol nor its security and enables us to decrease the size of the public key (less 17% for 128
bits security). We provide a new set of parameters denoted by HQC-RMRS for this decoding variation.
> 
> - Our implementations are now implemented in a constant-time way whenever relevant and they should not   leak any sensitive information with respect to timing attacks.
> 
> - Our implementations no longer rely on third party librairies for finite field arithmetic
>
> 
> The new implementations, documentation and KATs are directly available at:
> - Site: https://pqc-hqc.org
> - Specs:https://pqc-hqc.org/doc/hqc-specification_2020-04-21.pdf
> - Package :https://pqc-hqc.org/doc/hqc-submission_2020-04-21.zip

jschanck:
Will resolve #291 
Built using https://github.com/jschanck/package-pqclean

This needs a bit of work still. Maybe upstream developers should get involved. Patches directory in my scripts repo includes fixes for some significant problems, including some buffer overflows.

* Optional:


jschanck:
Any idea what version of clang the amd64-clang build is using? I can't reproduce the failure locally.

dstebila:
> Any idea what version of clang the amd64-clang build is using? I can't reproduce the failure locally.

https://github.com/PQClean/ci-containers

jschanck:
Another timeout on this build.

thomwiggers:
I've restarted CI

thomwiggers (approved):
Many thanks!

PR #325, opened 2020-09-14
Ban boolean operators in arithmetic expressions

jschanck:
Compilers will occasionally generate non-constant time code when boolean operators are used in arithmetic expressions.

For example, prior to Aug 23, 2020, there was 0/1 flag in a polynomial inversion routine in NTRU that was flipped using "!x" instead of "1-x". Recent versions of clang (but not gcc) with -O3 on haswell would generate two separate vectorized implementations of a particular loop, one for each value of x, and make a conditional jump. This jump could, in principle, leak the value of x. The bug was caught by TIMECOP.

The test_boolean.py script added by this PR catches the same bug. It also identifies several potential problems in other pqclean implementations. The issues with hqc are fixed in the more recent release. The issues in mceliece and ntru are false positives. The issues in rainbow are also (as far as I can tell) false positives. The issues in falcon need to be checked carefully.

### DIFF
--- a/crypto_kem/frodokem1344aes/clean/util.c
+++ b/crypto_kem/frodokem1344aes/clean/util.c
@@ -12,8 +12,10 @@
 #include "params.h"
 
 static inline uint8_t min(uint8_t x, uint8_t y) {
-  if (x < y) return x;
-  return y;
+    if (x < y) {
+        return x;
+    }
+    return y;
 }
 
 uint16_t PQCLEAN_FRODOKEM1344AES_CLEAN_LE_TO_UINT16(uint16_t n) {
--- a/crypto_kem/frodokem1344aes/clean/util.c
+++ b/crypto_kem/frodokem1344aes/clean/util.c
@@ -11,7 +11,10 @@
 #include "common.h"
 #include "params.h"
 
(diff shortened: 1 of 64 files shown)


thomwiggers:
Thanks for contributing this test. We do need to figure out how to handle these schemes that currently fail the test. Do we want to add some "skip the next line" functionality like many linters have? (I'm not sure comments survive the preprocessing that pycparser does) 

Probably we should rather fix these issues — but it does seem tricky in some cases...

jschanck:
Wasn't so bad. This run willl fail on hqc so this PR depends on #324.

thomwiggers:
So this needs to be rebased now and then we're good?

jschanck:
Hopefully! As you can tell from the Wconversion commits --- I did not run all of the tests locally.

thomwiggers:
It seems your changes to Falcon are not appreciated by the compiler...

jschanck:
Ugh, an error for negating an int16_t and storing it in an int16_t. Brilliant.

jschanck:
Oops. My version of fpr_lt is not correct. I thought about it for a bit, but I don't have a fix.

jschanck:
OK, this is ready for review now.

thomwiggers (approved):

PR #326, opened 2020-09-22
Add GeMSS

### LINKED ISSUE #139: Add Round 3 alternative signature scheme GeMSS

jschanck:
Will resolve #139 

* Optional:


jschanck:
I'm not sure how to proceed here without dynamic memory. The public keys for the 256 bit parameter sets are ~3MB, and this implementation ends up needing around 18 MB of stack space. Tests are failing because of the standard 8MB limit. The function changeVariablesMQS_simd_gf2, which computes a matrix map Q -> S*Q*S^t, is the main optimization target.

thomwiggers:
Not entirely unexpected that we would run into this at some point down the line. OQS has had stack-space trouble in the past already... 

How much stack is used if the pubkeys are moved into `malloc()`'d memory in the tests?

jschanck:
For gemss-256, we have
> nistkat (~3MB)
>> sign_keypairHFE (6.7MB)
>>> changeVariablesMQS_gf2 (7.2MB)

I can probably cut the usage of the two inner routines in half. We'd still need to switch to malloc in the tests.

thomwiggers:
CI has been refactored to build on Github Actions. Would you mind rebasing this PR on master and generating the workflow files for Github Actions (see `.github/workflows/generate_workflows.py`)? Sorry for the inconvenience.

jschanck:
The 128- and 192-bit parameter sets are good to go once we get a LICENSE file. I've reached out to two members of the GeMSS team about it and haven't heard back.

thomwiggers:
Have you heard back about the license?

jschanck:
I have not, and I'm not going to make any further attempts. Suggest closing this unless someone else wants to take over.

thomwiggers:
I've reached out to Ludovic Perret.

thomwiggers:
I've bumped the email

PR #333, opened 2020-10-05
ntru: packaging script update. resolves #331

### LINKED ISSUE #331: Use int16_t in both_negative_mask
The comment on `both_negative_mask` is wrong. It's only true for 16-bit arguments (`int16_t`) but wrong if `int` is larger (as it is on most systems). This isn't an issue here because the function should only be called with values of at most 16 bit length.
I would recommend using `int16_t` everywhere here to make the function do what the comment says.

https://github.com/PQClean/PQClean/blob/5eed5d3af1b3b6218fa4314cabc9510219600893/crypto_kem/ntruhrss701/clean/poly_r2_inv.c#L5-L8

jschanck:


thomwiggers (approved):
:+1:

PR #335, opened 2020-10-08
Fix dilithium namespacing issues

jschanck:
Fix namespacing issues on macOS/AVX2.


jschanck:
Alternative to this fix: allow function xxx to be namespaced as __NAMESPACE_xxx in test_symbol_namespace.py.

thomwiggers:
> Alternative to this fix: allow function xxx to be namespaced as __NAMESPACE_xxx in test_symbol_namespace.py.

This is actually a good point. Allowing `_NAMESPACE_xxx` on Linux in the first place is what caused this issue: because MacOS added some namespaces on top of the normal NS we are in trouble.

jschanck:
On second thought, symbols of the form `_[A-Z_].*` are reserved in C99. So removing the underscores in Dilithium (as this PR does) is the right approach.

thomwiggers:
CI has been refactored to build on Github Actions. Would you mind rebasing this PR on master? Sorry for the inconvenience.

thomwiggers (approved):

PR #344, opened 2020-10-21
Add AVX2 Falcon

jschanck:
New script for packaging Falcon at https://github.com/jschanck/package-pqclean.

There are some other implementations available in supercop. I could package the "tree" variants as well, which do some additional pre-computation in keygen.


jschanck:
Duplicate consistency is failing because of how astyle aligns argument lists that are split over multiple lines. This has come up before and I've just gone w/ long lines for argument lists. But maybe we should do something smarter in duplicate consistency. The simplest fix is to strip all whitespace before comparing files.

thomwiggers:
> Duplicate consistency is failing because of how astyle aligns argument lists that are split over multiple lines. This has come up before and I've just gone w/ long lines for argument lists. But maybe we should do something smarter in duplicate consistency. The simplest fix is to strip all whitespace before comparing files.

Yeah, this is something that's been coming up. Do you want to suggest a patch for this?

crypto_sign/falcon-1024/avx2/LICENSE
```diff
@@ -0,0 +1,24 @@
+\
```
thomwiggers, line 1: Is this intentional?

thomwiggers (approved):

PR #332, opened 2020-10-05
HQC: updated packaging script. should resolve #327

jschanck:
- remove duplicated includes.
- resolve `gen_matrix` issue by declaring static (see #327, #329)
- resolve shadowing of `vect_mask`

PR #345, opened 2020-10-23
hqc: Add -Wshadow and fix shadowing warnings

jschanck:
Resolves https://github.com/PQClean/PQClean/pull/332#issuecomment-714829999.

Just a note: adding "pclmul" flag to META.yml (1302bbbf) prevented the test suite from trying to build hqc on my development machine (even though it builds correctly there). On that machine, get_cpu_info() includes "pclmulqdq" but not "pclmul".

### DIFF
--- a/crypto_kem/hqc-128/META.yml
+++ b/crypto_kem/hqc-128/META.yml
@@ -22,9 +22,9 @@ principal-submitters:
   - Loïc Bidoux
 implementations:
     - name: clean
-      version: hqc-submission_2020-05-29 via https://github.com/jschanck/package-pqclean/tree/09ab89ed/hqc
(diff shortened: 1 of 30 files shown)


thomwiggers:
> Just a note: adding "pclmul" flag to META.yml (1302bbb) prevented the test suite from trying to build hqc on my development machine (even though it builds correctly there). On that machine, get_cpu_info() includes "pclmulqdq" but not "pclmul".


You're right, I have the same. I was sloppy. Would you mind adding that to this PR?

dstebila (dismissed):

thomwiggers (dismissed):

jschanck:
Hmm.. those are concerning errors. I'll look into it. Upstream is replacing hqc-{128,192,256} with hqc-rmrs-{128,192,256} for round 3, so if the error only affects hqc-{128,192,256} then it might not be worth fixing.

PR #348, opened 2020-10-26
Round 3 HQC

jschanck:
Upstream has removed the second round hqc-{128,192,256} parameter sets and replaced them with the second round hqc-rmrs-{128,192,256} parameter sets. I've left "rmrs-" in the name for now.

This also fixes the issues mentioned in #345.


thomwiggers (approved):
lgtm

PR #349, opened 2020-10-27
Round 3 Kyber

jschanck:
Round 3 Kyber. Packaged using scripts at https://github.com/jschanck/package-pqclean

The first commit, d7188f1, largely reproduces Thom's work in #307 (which looks good and should be merged IMO).
There are a few small differences here. Notably cdecl.inc -> cdecl.h and use of X and _X labels in .S files.


jschanck:
Upstream's aes256ctr.h is no longer compatible with our aes.h. I haven't looked too closely, but we might want to factor out the Kyber-specific parts and use a generic AES where possible.

thomwiggers:
We were already using a more specific one that uses our `aes.h`.

thomwiggers (approved):

PR #350, opened 2020-11-05
NTRU: faster multiplication for reference implementations

jschanck:
This replaces the reference code's schoolbook multiplication with Toom-4 + two layers of Karatsuba. I've also removed some unnecessary reductions from the AVX2 code, so that will get a bit faster too.

### DIFF
--- a/crypto_kem/ntruhps2048509/META.yml
+++ b/crypto_kem/ntruhps2048509/META.yml
@@ -23,9 +23,9 @@ auxiliary-submitters:
   - Zhenfei Zhang
 implementations:
     - name: clean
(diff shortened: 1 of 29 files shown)


jschanck:
I've seen this bug in clang-tidy a few times now. I think we should disable "clang-analyzer-valist.Uninitialized".

thomwiggers:
> I've seen this bug in clang-tidy a few times now. I think we should disable "clang-analyzer-valist.Uninitialized".

I've created #351. Meanwhile I've restarted the failed job on Travis.

thomwiggers (approved):

PR #352, opened 2020-11-09
Use valgrind to check for non-constant-time behavior (draft)

### LINKED ISSUE #12: Checking branching and memory access based on secret data
One of the requirements is that code should not branch based on secret data, and, similarly, not do memory accesses based on secret data.

A first intuition may be to pass an uninitialized secret key to the API functions and check that valgrind does not output any `[..] depends on uninitialised value(s)` errors. This has (at least) two problems:

- the secret key typically contains (part of) the public key as well; labeling all of it as secret is excessive
- rejection sampling based on secret data is not necessarily bad, e.g., sampling from the output of a XOF based on a secret seed

An alternative approach would be to have two `randombytes` functions, i.e., something like `randombytes_public` and `randombytes` (to error on the side of caution, I'd default to the assumed-secret variant). Normally they would both link to the same function, but when running this check the secret variant would not initialize the buffer.

This still leaves open the rejection sampling issue. It seems there is no way around manual inspection in these cases. We could whitelist the lines in which these issues occur, and then filter them out of the valgrind output.

Valgrind has support for suppression files, but the suppression rules seem too generic for our purposes. For example, consider [the rejection sampling in Kyber](https://github.com/PQClean/PQClean/blob/4cac434f47f5e2c99839cf8ea4f3a4241f5eb97c/crypto_kem/kyber768/clean/indcpa.c#L162), which triggers the following valgrind error and corresponding suppression rule

```
==6102== Conditional jump or move depends on uninitialised value(s)
==6102==    at 0x11439B: PQCLEAN_KYBER768_gen_matrix (indcpa.c:162)
==6102==    by 0x114769: PQCLEAN_KYBER768_indcpa_enc (indcpa.c:254)
==6102==    by 0x114D2D: PQCLEAN_KYBER768_crypto_kem_dec (kem.c:110)
==6102==    by 0x116EAC: test_null_sk (null_secretkey.c:32)
==6102==    by 0x116F10: main (null_secretkey.c:44)
==6102== 
{
   <insert_a_suppression_name_here>
   Memcheck:Cond
   fun:PQCLEAN_KYBER768_gen_matrix
   fun:PQCLEAN_KYBER768_indcpa_enc
   fun:PQCLEAN_KYBER768_crypto_kem_dec
   fun:test_null_sk
   fun:main
}
```

We could have a file with all whitelisted files and lines, but that quickly becomes a hassle to keep up to date when the line numbers change.

We could also include inline suppression markers (similar to `#noqa`), and then use this to filter valgrind output. Such markers are perhaps easily overlooked when doing code review, so we would probably want to search for these automatically and highlight them as part of the CI output (through a webhook and a comment on the PR?).

@mkannwischer and I briefly discussed this, but figured it would be worthwhile to hear some more thoughts on this before implementing anything concrete.

jschanck:
This PR adds a test that uses Valgrind to look for secret-information dependent branching and memory accesses. It potentially resolves #12 .

I've mostly followed SUPERCOP's implementation [[1]](http://bench.cr.yp.to/tips.html#timecop) of Moritz Neikes TIMECOP [[2]](https://post-apocalyptic-crypto.org/timecop/index.html). 

Each call to randombytes(buf, len) is intercepted and the len bytes at buf are marked as undefined using Valgrind's "VALGRIND_MAKE_MEM_UNDEFINED" [[3]](https://www.valgrind.org/docs/manual/mc-manual.html#mc-manual.clientreqs). Any memory that depends on this undefined memory will also be marked as undefined when the program is run through Valgrind's memcheck tool. Valgrind outputs an error if the program branches on undefined memory.

This produces some false positives. Unlike SUPERCOP and TIMECOP, I've made it so that we can flag false positives with a comment. For example, Kyber does some rejection sampling while generating the public `A` matrix. On Line [124 of indcpa.c](https://github.com/jschanck/PQClean/blob/ed8ddd7daa1ebb157c5e79750a8a0a8db3a951de/crypto_kem/kyber768/clean/indcpa.c#L124) I've added:
```        //+PQCLEAN_TIMECOP_DECLASSIFY(&val, sizeof val);```
The test_timecop.py script strips the "//+" prefix from this before compiling the program and running it through memcheck.

Some TODOs:
- Flag false positives. I've done the clean implementations of dilithium3 and kyber768 as examples.
- Flesh out test/crypto_{kem,sign}/timecop.c. We should check that failure branches are constant time We might want to test on actual random inputs, which means changing test/common/poisonbytes.c.

### DIFF
--- a/crypto_kem/kyber768/clean/indcpa.c
+++ b/crypto_kem/kyber768/clean/indcpa.c
@@ -121,6 +121,7 @@ static size_t rej_uniform(int16_t *r, size_t len, const uint8_t *buf, size_t buf
         val = (uint16_t)(buf[pos] | ((uint16_t)buf[pos + 1] << 8));
         pos += 2;
 
+        //+PQCLEAN_TIMECOP_DECLASSIFY(&val, sizeof val);
         if (val < 19 * KYBER_Q) {
             val -= (uint16_t)((val >> 12) * KYBER_Q); // Barrett reduction
             r[ctr++] = (int16_t)val;
--- a/crypto_sign/dilithium3/clean/poly.c
+++ b/crypto_sign/dilithium3/clean/poly.c
@@ -234,6 +234,7 @@ int PQCLEAN_DILITHIUM3_CLEAN_poly_chknorm(const poly *a, uint32_t B) {
         t ^= (t >> 31);
         t = (Q - 1) / 2 - t;
 
+        //+PQCLEAN_TIMECOP_DECLASSIFY(&t, sizeof t);
         if ((uint32_t)t >= B) {
             return 1;
         }
@@ -270,6 +271,7 @@ static size_t rej_uniform(
         t |= (uint32_t)buf[pos++] << 16;
         t &= 0x7FFFFF;
 
+        //+PQCLEAN_TIMECOP_DECLASSIFY(&t, sizeof t);
         if (t < Q) {
             a[ctr++] = t;
         }
@@ -345,6 +347,8 @@ static size_t rej_eta(
         t0 = buf[pos] & 0x0F;
         t1 = buf[pos++] >> 4;
 
+        //+PQCLEAN_TIMECOP_DECLASSIFY(&t0, sizeof t0);
+        //+PQCLEAN_TIMECOP_DECLASSIFY(&t1, sizeof t1);
         if (t0 <= 2 * ETA) {
             a[ctr++] = Q + ETA - t0;
         }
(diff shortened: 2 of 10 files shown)


jschanck:
Do the valgrind tests not run in CI? I just checked the round3 kyber PR and I didn't see test_valgrind running there.

thomwiggers:
Yeah, the sanitizer and valgrind tests are not run because they take forever (notably, but not limited to, SPHINCS+) When they were disabled, we were not splitting up everything though, so we should probably try if we can enable these tests again.

PR #353, opened 2020-11-10
Disable skipping slow tests

thomwiggers:
See #352. We were skipping slow tests because they made us exceed run times,
but we are no longer as limited as we now split runtime for each individual
scheme.


thomwiggers:
``` 
Command being timed: "../bin/functest_sphincs-sha256-192s-robust_clean"                           
[..]
        Maximum resident set size (kbytes): 14250608                                                      
```

How do I download more RAM to CI?

Issue #355, opened 2020-11-16
Falcon code fails "strict" compilation


mouse07410:
Lines
https://github.com/PQClean/PQClean/blob/3c8be3cb1f8ca0953f2df44ea665f118efb468d6/crypto_sign/falcon-1024/avx2/rng.c#L94
and
https://github.com/PQClean/PQClean/blob/3c8be3cb1f8ca0953f2df44ea665f118efb468d6/crypto_sign/falcon-1024/avx2/rng.c#L99
cause the GCC compiler to emit warning "`dereferencing type-punned pointer will break strict-aliasing rules`".

As we compile with `-Werror`, it breaks our build. This applies to both `avx2` and `clean` versions.

One possible fix is replacing those with something like `memcpy(&cc, (p->state.d + 48), 8);`

jschanck:
To avoid the strict-aliasing warning we should access the memory with pointers that are declared as part of the unions. So we should change

https://github.com/PQClean/PQClean/blob/3c8be3cb1f8ca0953f2df44ea665f118efb468d6/crypto_sign/falcon-1024/avx2/inner.h#L462-L473

to

```c
typedef struct {
    union {
        uint8_t d[512]; /* MUST be 512, exactly */
        uint32_t d32[512/4];
        uint64_t d64[512/8];
    } buf;
    size_t ptr;
    union {
        uint8_t d[256];
        uint32_t d32[256/4];
        uint64_t d64[256/8];
    } state;
    int type;
} prng;
```

The alignment trick that is causing the warning is common throughout the PQClean codebase. So I think we should start by developing a test that catches all instances of this.

@mouse07410 What version of GCC are you using? And does it issue aliasing warnings in any schemes other than falcon?

mouse07410:
> What version of GCC are you using? 

```
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/lto-wrapper
Target: x86_64-redhat-linux
[...]

PR #358, opened 2020-11-18
hqc-rmrs-256: remove unused static vars

jschanck:
Remove two unused static variables. Fixes part of https://github.com/PQClean/PQClean/issues/355#issuecomment-729970705


thomwiggers:
Weird that they did not get removed by the compiler or warned on.

thomwiggers (approved):

PR #367, opened 2020-12-11
FrodoKEM: Fix bug in the output of the ct_verify function

mkannwischer:
A bug in the CCA transformation was reported on the pqc-forum on 2020-12-10:
https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/kSUKzDNc5ME

It was fixed today in https://github.com/microsoft/PQCrypto-LWEKE/commit/669522db63850fa64d1a24a47e138e80a59349db.
This PR ports that fix to PQClean.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -24,6 +24,9 @@ Use at your own risk.
 ### 2019-XX-XX
 -->
 
+### 2020-12-11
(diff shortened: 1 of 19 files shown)


dstebila (approved):

dstebila (changes requested):
Don't forget to update the URL in the META.yml files.

dstebila (approved):

PR #381, opened 2021-03-12
HQC: required flag is called pclmul not pclmulqdq

jschanck:
The flag probably _should_ be called pclmulqdq, but it is not. Liboqs was not building the AVX2 implementation of HQC because of this.

### DIFF
--- a/crypto_kem/hqc-rmrs-128/META.yml
+++ b/crypto_kem/hqc-rmrs-128/META.yml
@@ -33,4 +33,4 @@ implementations:
             required_flags:
                 - avx2
                 - bmi1
-                - pclmulqdq
+                - pclmul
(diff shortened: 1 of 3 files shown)


thomwiggers:
`required_flags` maps the CPU features, not the compiler flags that are used. `/proc/cpuinfo` says the flag is `pclmulqdq` on my i7-8550U.

thomwiggers:
the `bmi1` flag has similar issues, and is [handled by liboqs e.g. here](https://github.com/open-quantum-safe/liboqs/blob/fc35e09a1491fe3ffaba90999c1767e65a694ac9/scripts/copy_from_upstream/src/kem/family/CMakeLists.txt#L15).

PR #302, opened 2020-06-12
Script for generating an export to SUPERCOP

thomwiggers:
Also updates some type annotations in the helpers

PR #382, opened 2021-03-15
Fix McEliece sse/avx for use in shared libraries

jschanck:
The McEliece sse and avx const.S files declare some constants with global visibility, but the hand written assembly doesn't handle relocation correctly. This caused an issue in liboqs https://github.com/open-quantum-safe/liboqs/discussions/930.

The symbols are only needed inside McEliece. This PR marks them .hidden on linux or .private_extern on macOS.


thomwiggers (approved):

thomwiggers:
Thanks. 

There are probably other problems — I've run into issues when I exported the code from PQClean into supercop (#302): stuff segfaulted. The McEliece code should maybe be updated from upstream, but the method of releases is _suboptimal_ to figure out what changed exactly 😞

jschanck:
The only other issue that I've run into is the use of the __m256i_u type, eg https://github.com/PQClean/PQClean/blob/dbbfe0dfc50655b7b34b37d85fbe68b2fd4452ba/crypto_kem/mceliece348864/avx/int32_sort.c#L7
This is a fairly recent gcc invention. It's made its way into clang as of clang-9. But I don't think we should be using it. There's a chance that gcc complains about __m256i here. My preference would be to use __mm256i and suppress the warning.

thomwiggers:
Yeah that's an issue we've been facing/blissfully ignoring for a while now.

The problem is that we want the PQClean code to compile with all the sensible warnings turned on, just copy/pasting code from a folder into your own project without thinking too much. That's obviously what enables having scripts such as the `copy_from_pqclean.sh`.  Having to figure out what compiler warnings you have to turn off per scheme is just messy — and you need to do that for each individual compiler because AFAIK there's no way to disable them inline (cf. `#[allow(warning)]` in Rust). 

I'm thinking more and more we should relax our `#ifdef` requirements. Perhaps it would be a good idea to have a `compat.h` file in `common`; this file could contain 

```
#ifdef GCC_VERSION >= 9
  typedef _m256i_u _m256i;
#endif 
```

thomwiggers:
oh and;
> there's a chance that gcc complains about __m256i here.

It definitely does; I had to make all those adjustments for `__m256i_u` myself.

jschanck:
OK, I won't try to tack on a fix to the __m256i_u issue here. This is ready to go.

compat.h sounds like a good idea. Maybe open an issue so we can discuss other things that might go in it?

thomwiggers:
I've created issue #383

Issue #383, opened 2021-03-16
Irreconcilable compiler differences


thomwiggers:
We keep running into differences between compilers and versions of compilers that make it hard to write code that a) just universally compiles and b) doesn't raise any compliants.

> > ... about problems with `_m256i` having alignment requirements but `_m256i_u` not existing in certain versions of gcc / clang ...
> The problem is that we want the PQClean code to compile with all the sensible warnings turned on, just copy/pasting code from a folder into your own project without thinking too much. That's obviously what enables having scripts such as the `copy_from_pqclean.sh`.  Having to figure out what compiler warnings you have to turn off per scheme is just messy — and you need to do that for each individual compiler because AFAIK there's no way to disable them inline (cf. `#[allow(warning)]` in Rust). 

I'm thinking more and more we should relax our `#ifdef` requirements. Perhaps it would be a good idea to have a `compat.h` file in `common`; this file could contain 

```
#if GCC_VERSION < 9
  typedef _m256i_u _m256i;
#endif 
```

_See also https://github.com/PQClean/PQClean/issues/382#issuecomment-800085270_

thomwiggers:
A similar recurring issue is the need for a macro that handles appending `_` to symbols used in assembly files.

thomwiggers:
Along these lines: an `align.h` *with ifdefs and compiler-specific attributes* may also be interesting to clean up the alignment mess...

PR #385, opened 2021-03-19
Compatibility shim for unaligned vector types

thomwiggers:
This introduces an intial version of the compatibility shim `compat.h` which will help to provide compatibility in schemes without copying too much code or worrying about compiler versions on a scheme-level.

See #383

It also patches the Classic McEliece issue mentioned in #382

### DIFF
--- a/common/compat.h
+++ b/common/compat.h
@@ -1,0 +1,47 @@
+#ifndef  PQCLEAN_COMMON_COMPAT_H
+#define PQCLEAN_COMMON_COMPAT_H
+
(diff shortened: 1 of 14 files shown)


thomwiggers:
FYI: @jschanck

thomwiggers:
I tested out this code on godbolt.org; it seems to work great. The only compiler that complained was `icc`; but until Intel pays me for an icc license and for my hours I'm not inclined to invest much in that 😄

PR #387, opened 2021-03-26
LEDAcrypt submission

Bob10918:
Submission of post-quantum crypto algorithm LEDAcrypt (www.ledacrypt.org). The submission on PQClean is part of a master thesis project. 
Actually only the first implementation with a specific set of parameters is present; the others will come shortly using generated code from a shared code base.

* Optional:


crypto_kem/ledacrypt23371/clean/niederreiter.h
```diff
@@ -0,0 +1,32 @@
+#ifndef NIEDERREITER_H
+#define NIEDERREITER_H
+
+#include "qc_ldpc_parameters.h"
+#include "gf2x_limbs.h"
+#include "gf2x_arith_mod_xPplusOne.h"
+
+
+/*----------------------------------------------------------------------------*/
+#pragma pack(1)
```
thomwiggers, line 10: This isn't portable and should be removed.

crypto_kem/ledacrypt23371/clean/niederreiter.h
```diff
@@ -0,0 +1,32 @@
...
+
+typedef struct {
+   DIGIT Mtr[(N0-1)*NUM_DIGITS_GF2X_ELEMENT];
+   // Dense representation of the matrix M=Ln0*L,
+   // An array including a sequence of (N0-1) gf2x elements;
+   // each gf2x element is stored as a binary polynomial(mod x^P+1)
+   // with P coefficients.
+} publicKeyNiederreiter_t;
+
+#pragma pack()
```
thomwiggers, line 29: `#pragma` should not be used.

crypto_kem/ledacrypt23371/clean/niederreiter_encrypt.c
```diff
@@ -0,0 +1,107 @@
...
+   AES_XOF_struct hashedAndTruncatedSeed_expander;
+   memset(&hashedAndTruncatedSeed_expander, 0x00, sizeof(AES_XOF_struct));
+    PQCLEAN_LEDACRYPT23371_CLEAN_seedexpander_from_trng(&hashedAndTruncatedSeed_expander,
+                                                        hashedAndTruncatedSeed);
+
+   POSITION_T errorPos[NUM_ERRORS_T];
+    PQCLEAN_LEDACRYPT23371_CLEAN_rand_error_pos(errorPos, &hashedAndTruncatedSeed_expander);
+
+   DIGIT error_vector[N0*NUM_DIGITS_GF2X_ELEMENT];
+    PQCLEAN_LEDACRYPT23371_CLEAN_expand_error(error_vector, errorPos);
```
thomwiggers, line 81: sloppy indentation, it's similarly messy in many places. I'm not sure why `astyle` doesn't try to fix this...

crypto_kem/ledacrypt23371/clean/niederreiter_keygen.c
```diff
@@ -0,0 +1,76 @@
+#include "niederreiter_keygen.h"
+#include "H_Q_matrices_generation.h"
+#include "gf2x_arith_mod_xPplusOne.h"
+#include "rng.h"
+#include "dfr_test.h"
+#include <string.h>
+#include "randombytes.h"
+#include "nistseedexpander.h"
+/*----------------------------------------------------------------------------*/
+/* Implementation that should never be optimized out by the compiler */
```
thomwiggers, line 10: Why is this assumption necessary? Even volatile would not guarantee that this does not happen. In general, preventing something isn't optimised out can't be guaranteed in C.

crypto_kem/ledacrypt23371/clean/rng.c
```diff
@@ -0,0 +1,280 @@
...
+
+void PQCLEAN_LEDACRYPT23371_CLEAN_initialize_pseudo_random_generator_seed(int seedFixed, char *seed)
+{
+
+   if (seedFixed == 1)
+      srand(atoi(seed));
+   else {
+      uint8_t r[4];
+      randombytes(r, 4);
+      srand(*r);
```
thomwiggers, line 30: **`srand`**? What is an unsafe RNG doing in crypto code?

crypto_kem/ledacrypt23371/clean/rng.c
```diff
@@ -0,0 +1,280 @@
...
+#include "nistseedexpander.h"
+
+
+/******************************************************************************/
+/*----------------------------------------------------------------------------*/
+/*              start PSEUDO-RAND GENERATOR ROUTINES for rnd.h                */
+/*----------------------------------------------------------------------------*/
+
+
+void PQCLEAN_LEDACRYPT23371_CLEAN_initialize_pseudo_random_generator_seed(int seedFixed, char *seed)
```
thomwiggers, line 22: What is this function necessary for?

crypto_kem/ledacrypt23371/clean/rng.h
```diff
@@ -0,0 +1,72 @@
+#ifndef RNG_H
+#define RNG_H
+
+/******  From this point on, the code was supplied by NIST ****************/
```
thomwiggers, line 4: Use the PQClean API

crypto_kem/ledacrypt23371/clean/bf_decoding.c
```diff
@@ -0,0 +1,65 @@
...
+   int iteration = 0;
+
+   do {
+      gf2x_copy(currSyndrome, privateSyndrome);
+      memset(unsatParityChecks,0x00,N0*P*sizeof(uint8_t));
+      for (int i = 0; i < N0; i++) {
+         for (int valueIdx = 0; valueIdx < P; valueIdx++) {
+            for(int HtrOneIdx = 0; HtrOneIdx < V; HtrOneIdx++) {
+               POSITION_T tmp = (HtrPosOnes[i][HtrOneIdx]+valueIdx) >= P ?
+                                (HtrPosOnes[i][HtrOneIdx]+valueIdx) -P : (HtrPosOnes[i][HtrOneIdx]+valueIdx);
```
thomwiggers, line 31: There are many of these ternary if statements. Is the value of `HtrPosOnes[i][HtrOneIdx]+valueIdx` or that of `tmp` security-sensitive?

crypto_kem/ledacrypt23371/clean/bf_decoding.c
```diff
@@ -0,0 +1,65 @@
...
+               if (gf2x_get_coeff(currSyndrome, tmp))
+                  unsatParityChecks[i*P+valueIdx]++;
```
thomwiggers, line 33: Are you sure this does not leak any timing information?

thomwiggers (changes requested):
Thanks for your submission to PQClean, we understand that this took a lot of effort and we appreciate it a great deal.

## Inclusion of not-PQC-NIST-competition-candidates
The previous version of LEDAcrypt was removed from PQClean after it was removed from the NIST competition in Round 3. We've consistently dropped schemes from PQClean whenever they were dropped from the NIST competition — though that was motivated as much by the time it would save us.

That does mean we need to make a decision on whether to track schemes that are not (/no longer) part of the NIST competition. Before you put a lot more effort into adressing the below comments just for inclusion into PQClean, it may be wise to wait for us to reach a conclusion on this.

## Other remarks
I've also got some other remarks. Part are pointed out by comments I made on individual lines of code.

* The previous version of the code was not protected against timing attacks. I've found some sections of code that still heavily use `if` statements, either explicitly or through ternary ifs `a ? b : c`. There's also a bunch of non-trivial array indexing that happen, which may leak time as well. I will readily admit that I haven't read the LEDA specification _at all_, so I don't know if e.g. `bf_decoding.c` is even to be considered sensitive for timing leakages, but unless it's a part of rejection sampling, it could be a problem. I'd appreciate it if you could comment if these instances I've found are okay, and maybe double-check the remainder of the (hidden) `if` statements and any remaining lookup tables.

* Speaking of lookup tables, the included AES-implementation (which should be replaced by our API) is completely insecure against timing side-channels...

* It seems testing code is included (the rng functions). Those should probably be removed?

* You need to generate tests through the script in `.github/workflows` as well.

crypto_kem/ledacrypt23371/clean/LICENSE.txt
```diff
@@ -0,0 +1,30 @@
+/** 
+ *
+ * @version 3.0 (April 2020)
+ *
+ * ISO-C11 compliant implementations of LEDAcrypt ciphers 
+ * (LEDAcrypt-KEM, LEDAcrypt-KEM-CPA, LEDAcrypt-PKC) using GCC built-ins.
```
thomwiggers, line 6: ISO C with GCC builtins seems a contradictio in terminis...

thomwiggers:
As per the discussion in #388 we will unfortunately not be accepting this pull request.  I fully understand how much work must've gone into it, and we do appreciate it — it's just that we can beter focus on other things here.

PR #391, opened 2021-04-08
Run Linux jobs on own runners

thomwiggers:


thomwiggers:
self-hosted runners are.... not great.

https://github.com/actions/runner/issues/434

thomwiggers:
Not worth the hassle.

PR #393, opened 2021-05-05
HQC-192/avx2: fix missing initialization in compute_syndromes

jschanck:
Resolves https://github.com/open-quantum-safe/liboqs/issues/990


jschanck:
@thomwiggers good to go

thomwiggers:
As I've seen problems in the 128-bit version as well, I'll try to do some testing.

thomwiggers (approved):

PR #395, opened 2021-06-02
Pull latest NTRU Prime from supercop-20210529

jschanck:
Integrate the latest NTRU Prime code. Upstream has made some performance improvements, but the code is interoperable with previous releases and does not add any parameter sets (although other parameter sets are available upstream).

I've removed an unnecessary AES shim that my packaging script was writing to the sntrup directories.

Known issues: clang-tidy doesn't like the use of recursion in Fq_pow and int32_sort. I can add NOLINT annotations, but I'd actually prefer to disable the warning.


crypto_kem/ntrulpr761/clean/kem.c
```diff
@@ -156,7 +158,9 @@ static void Hide(unsigned char *c, unsigned char *r_enc, const Inputs r, const u
             s[0] = 5;
             Hash(h, s, sizeof s);
         }
-        PQCLEAN_NTRULPR761_CLEAN_crypto_stream_aes256ctr((unsigned char *) L, 4 * p, aes_nonce, h);
+        if (PQCLEAN_NTRULPR761_CLEAN_crypto_stream_aes256ctr((unsigned char *) L, 4 * p, aes_nonce, h) != 0) {
+            abort();
+        }
```
thomwiggers, line 163: could this `abort()` not better be handled in the function that wraps `aes256ctr`?

thomwiggers (commented):
Thanks for the update.

I've got one small question.

crypto_kem/ntrulpr761/clean/kem.c
jschanck: Thanks. Upstream might call OpenSSL's aes directly at that line, but the abort isn't necessary with the wrapper. I've removed it.

thomwiggers (approved):

dstebila:
This code has now propagated to the OQS benchmarking and @baentsch has observed an improvement in performance for NTRU Prime decaps but a slow down in performance for other aspects of NTRU Prime.  Is that expected?

baentsch:
Here [a link to the performance numbers before](https://openquantumsafe.org/benchmarking/visualization/2021-07-14/speed_kem.json) and [here a link to new numbers after this new code got added](https://openquantumsafe.org/benchmarking/visualization/2021-07-15/speed_kem.json).

thomwiggers:
Those JSON dumps weren't very accessible, so here I've at least pretty-printed them: https://gist.github.com/thomwiggers/a85c372fdef9d272614aca20f1993a79

PR #409, opened 2021-07-30
Remove non-ASCII characters from Falcon comments

dstebila:
Some comments in the Falcon source code use UTF-8 characters. This causes problems in some older platforms I was testing liboqs on. This PR replaces the UTF-8 characters with ASCII.


dstebila:
A bunch of failures that seem to be unrelated to the change I made. Not sure how you want to deal with that...

mkannwischer (approved):
LGTM

PR #410, opened 2021-08-03
Fix assertion error seen on aarch64_test 

thomwiggers:


thomwiggers:
@mkannwischer this should allow you to rebase your PR on top of this one (once it gets merged into `master`) which should then in turn not spawn a gazillion tests because a test/*.py file changed.

mkannwischer:
Looks good. 
According to https://github.com/workhorsy/py-cpuinfo, `cpuinfo['arch']` can be `"X86_32", "X86_64", "ARM_8", "ARM_7", "PPC_32", "PPC_64", "SPARC_32", "SPARC_64", "S390X", "MIPS_32", "MIPS_64"`. 
Does that mean that `x86` is also wrong?

mkannwischer (approved):

thomwiggers:
Apparently. It's not been a problem for our tests, as the CPU is detected there as x86_64 anyway  (but still as 32-bit mode) https://github.com/PQClean/PQClean/blob/master/test/pqclean.py#L171-L173

PR #414, opened 2021-08-12
Disable more clang-tidy checks

thomwiggers:
Makes clang-tidy-11 work


thomwiggers:
None of the failures are caused by this PR.

PR #416, opened 2021-09-07
Add aarch64 implemenations of Kyber

mkannwischer:
This is a WIP for integrating the Kyber implementations from https://eprint.iacr.org/2021/986. 

Once this is functional, we probably want to merge https://github.com/PQClean/PQClean/commit/037b8be42c680fc8f59d682583291cbff8c442c6 as a separate PR. 

* Optional:


mkannwischer:
Let me rebase this on top of #422. Then the errors should go away.

PR #422, opened 2021-09-23
Arm-v8 test framework changes

mkannwischer:
I'm working on adding Neon implementations for Kyber (#416) and later also Saber and Dilithium and I ran into a problem on aarch64: pycparser's fake stdlib does not support the neon headers which results in to errors like this: 
```
/home/mjk/git/PQClean/test/test_boolean.py:83: 
...
E               subprocess.CalledProcessError: Command '['cc', '-E', '-std=c99', '-nostdinc', '-I/home/ubuntu/PQClean/test/common', '-I/home/ubuntu/PQClean/test/../common', '-I/home/ubuntu/PQClean/test/pycparser/utils/fake_libc_include', '../crypto_kem/kyber768/aarch64/kem.c']' returned non-zero exit status 1.

/usr/lib/python3.9/subprocess.py:528: CalledProcessError
---------------------------------------------------- Captured stderr call -----------------------------------------------------
In file included from ../crypto_kem/kyber768/aarch64/symmetric.h:61,
                 from ../crypto_kem/kyber768/aarch64/kem.c:6:
../crypto_kem/kyber768/aarch64/fips202x2.h:5:10: fatal error: arm_neon.h: No such file or directory
    5 | #include <arm_neon.h>
      |          ^~~~~~~~~~
 ```
 
 I don't see any way around this, so I have disabled the tests using pycparser for implementations targetting arm_8. 

Additionally, valgrind was not enabled for aarch64 even though it works perfectly fine. I've enabled it.

### DIFF
--- a/test/test_char.py
+++ b/test/test_char.py
@@ -7,6 +7,7 @@ This is ambiguous; compilers can freely choose `signed` or `unsigned` char.
 import os
 
 import pytest
+import unittest
 
 import helpers
 import platform
@@ -48,7 +49,7 @@ def test_char(implementation):
     if 'supported_platforms' in implementation.metadata():
         for platform in implementation.metadata()['supported_platforms']:
             if platform['architecture'] == "arm_8":
-                return
+                raise unittest.SkipTest()
 
     for fname in os.listdir(implementation.path()):
         if not fname.endswith(".c"):
--- a/test/test_char.py
+++ b/test/test_char.py
@@ -9,6 +9,7 @@ import os
 import pytest
 
 import helpers
+import platform
 import pqclean
 import pycparser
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by thomwiggers** [comment]
You could consider adding the types and an `arm_neon.h` to https://github.com/eliben/pycparser/blob/master/utils/fake_libc_include/_fake_defines.h#L209

File: test/test_char.py [hunk 1]
```diff
@@ -43,6 +44,12 @@ def walk_tree(ast, parent=[]):
 @helpers.filtered_test
 def test_char(implementation):
     errors = []
+    # pyparser's fake_libc does not support the Arm headers
+    if 'supported_platforms' in implementation.metadata():
+        for platform in implementation.metadata()['supported_platforms']:
+            if platform['architecture'] == "arm_8":
+                return
```
thomwiggers, line 51: we've been doing `raise unittest.SkipTest()` instead of `return` but otherwise it looks good.

thomwiggers (changes requested):
Do you want to fix `pycparser` up with arm support or do you say "let's leave it".

test/test_char.py
mkannwischer: good catch, thanks!

thomwiggers (approved):

mkannwischer:
Forgot the import of unittest, sorry.

thomwiggers:
> Do you want to fix pycparser up with arm support or do you say "let's leave it".

No is a perfectly acceptable answer here ;-)

mkannwischer:
Let's leave it like this for now. If I find the time, I'll look into patching pycparser later.

thomwiggers (approved):

PR #425, opened 2021-10-04
Update randombytes

thomwiggers:
Updates randombytes from upstream and namespaces it.

### DIFF
--- a/common/randombytes.h
+++ b/common/randombytes.h
@@ -16,6 +16,7 @@ extern "C" {
 /*
  * Write `n` bytes of high quality random bytes to `buf`
  */
+#define randombytes     PQCLEAN_randombytes
 int randombytes(void *buf, size_t n);
 
 #ifdef __cplusplus
--- a/common/randombytes.h
+++ b/common/randombytes.h
@@ -1,13 +1,25 @@
-#ifndef RANDOMBYTES_H
-#define RANDOMBYTES_H
-#include <stdint.h>
+#ifndef sss_RANDOMBYTES_H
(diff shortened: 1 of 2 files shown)


common/randombytes.h
```diff
@@ -1,13 +1,26 @@
-#ifndef RANDOMBYTES_H
-#define RANDOMBYTES_H
-#include <stdint.h>
+#ifndef sss_RANDOMBYTES_H
```
mkannwischer, line 1: Why is this `sss_` now?

common/randombytes.h
```diff
@@ -1,13 +1,26 @@
...
-#endif
+#endif /* _WIN32 */
 
-int randombytes(uint8_t *buf, size_t n);
 
+/*
+ * Write `n` bytes of high quality random bytes to `buf`
+ */
+#define randombytes     PQCLEAN_randombytes
+int randombytes(void *buf, size_t n);
```
mkannwischer, line 20: This should be a `uint8_t *`?

mkannwischer (changes requested):

common/randombytes.h
thomwiggers: Because that's how I copied it from upstream, and how @dsprenkels had it. We could change it to `RANDOMBYTES_H` though.

common/randombytes.h
thomwiggers: Yeah I guess this is an API change otherwise.

thomwiggers:
@mkannwischer can we merge this?

PR #427, opened 2021-11-03
Add ntruhps40961229/clean and ntruhrss1373/clean

### LINKED ISSUE #399: Add NTRU Level 5 param set soonest please
Following NIST formal request, NTRU team released NIST Sec Level 5 parameter set for NTRU HPS and NTRU HRSS algorithms: https://github.com/jschanck/ntru 

Can you please add their implementations to the codebase the soonest?

Thanks!

jschanck:
This adds reference implementations of two level 5 NTRU parameter sets.

Resolves #399. This will also allow liboqs to resolve https://github.com/open-quantum-safe/liboqs/issues/1031


dstebila:
Need to add CI jobs in https://github.com/PQClean/PQClean/tree/master/.github/workflows to actually test the new algorithms.

thomwiggers:
I've pushed updated workflows. There should be a CI check that should fail in this scenario though...

dstebila:
There look to be to two types of CI failures.  

1. On the PPC tests it identifies some of the array indices as being out of bounds.  I think these need to be checked.
2. On the Windows tests, it complains about some if statements having constant expressions.  I think we could disable this warning since we don't mind such behaviour.

jschanck:
The PPC errors are documentation issues, not actual array out of bounds problems. It's on my todo list to fix.

dstebila:
> The PPC errors are documentation issues, not actual array out of bounds problems. It's on my todo list to fix.

I'd offer to do it, but I'm not sure what you mean by documentation issues.

crypto_kem/ntruhps40961229/META.yml
```diff
@@ -23,4 +23,14 @@ auxiliary-submitters:
...
+    - name: avx2
+      version: https://github.com/jschanck/ntru/tree/bc5e7efe avx2 implementation
+      supported_platforms:
+          - architecture: x86_64
+            operating_systems:
+                - Linux
+                - Darwin
+            required_flags:
+                - avx2
+                - bmi2
```
dstebila, line 36: Is this a mistake? I don't think the PR has an AVX2 implementation of NTRU HPS-4096-1229?

```suggestion
```

crypto_kem/ntruhrss1373/META.yml
```diff
@@ -23,4 +23,14 @@ auxiliary-submitters:
...
+    - name: avx2
+      version: https://github.com/jschanck/ntru/tree/bc5e7efe avx2 implementation
+      supported_platforms:
+          - architecture: x86_64
+            operating_systems:
+                - Linux
+                - Darwin
+            required_flags:
+                - avx2
+                - bmi2
```
dstebila, line 36: Ditto.

```suggestion
```

jschanck:
Thanks, Douglas. And to clarify the array parameters are "documentation issues" because the C standard says that the compiler has to demote an array parameter like `void f(int x[10])` to a pointer like `void f(int *x)`. So the only reason to put the length in the declaration is for documentation.

baentsch:
@jschanck @dstebila Thanks for moving this to "green". What's required/missing now to merge this? I'd gladly integrate it into `liboqs` and all downstreams then to move the 0.7.1 release forward.

PR #433, opened 2022-01-01
Update dilithium/{clean,avx}. Resolves #421

### LINKED ISSUE #421: Update Dilithium
We need to pull in the fixes for GCC 11 https://github.com/pq-crystals/dilithium/pull/53

jschanck:
Update Dilithium to pull in GCC 11 fixes.

Also note that I originally applied 52851284 directly in the PQClean tree instead of in my packaging scripts. I had hoped upstream would take it, but they fixed the UB with a `__may_alias__` extension instead. So now I've backported the fix to my packaging scripts. I changed the memmove to a memcpy because of https://github.com/open-quantum-safe/profiling/issues/47.

### DIFF
--- a/crypto_sign/dilithium2/META.yml
+++ b/crypto_sign/dilithium2/META.yml
@@ -17,9 +17,9 @@ auxiliary-submitters:
   - Damien Stehlé
 implementations:
     - name: clean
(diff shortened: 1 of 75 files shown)


thomwiggers:
It seems these tests are failing due to a new version of clang-tidy on PPC (which is based on Debian Unstable) (`altera-` lints) and an issue with the Github actions runners on MacOS.

thomwiggers:
I've rebased, let's see what happens.

thomwiggers:
The MacOS thing is #435.

thomwiggers (approved):

PR #437, opened 2022-02-18
Fix a bunch of clang-tidy 12 errors

thomwiggers:
I've added some more lints to the ignore lists because they gave stupid warnings, and fixed some issues in a few schemes.


thomwiggers:
Closes #398

dstebila (approved):

PR #438, opened 2022-02-21
CryptGenRandom expects DWORD

thomwiggers:
I accidentally missed this Windows 64-bit build failure.

cc @dsprenkels


dsprenkels:
@thomwiggers Thanks for the heads-up!

PR #444, opened 2022-04-21
crypto_kem/sntrup761: replace bitwise operators with intrinsics

matbech:
MSVC does not support AVX bitwise operators.

This replaces all those operators with the following intrinsics:
_mm256_and_si256
_mm256_or_si256
_mm256_xor_si256


thomwiggers:
Hi, thanks for your contribution.

The NTRU prime parametersets don't claim Windows support, but it seems you're right, this should be all that is needed. Can you also update [the meta file](https://github.com/PQClean/PQClean/blob/master/crypto_kem/sntrup761/META.yml), but even better, do you think you can integrate these changes via the patch set at https://github.com/mkannwischer/package-pqclean/tree/main/ntruprime/patches so that we can apply them to all parametersets?

thomwiggers (changes requested):
The metafile should be updated to indicate Windows as a supported platform; additionally we probably want to apply the same patches to all parametersets.

thomwiggers:
It seems there's also no NMake Makefile yet (which makes sense)

PR #449, opened 2022-07-18
Compile functests with GCC's static analyzer

### LINKED ISSUE #442: Integrate GCC static analyzer into PQClean
GCC 10+ has `-fanalyzer`

thomwiggers:
Compile the `functest` test with `-fanalyzer` on Linux only.

Closes #442


thomwiggers:
This fails on McEliece now

PR #455, opened 2022-08-08
Fix HQC C aliasing violation

### LINKED ISSUE #453: crypto_kem/hqc-rmrs-128/avx2/reed_muller.c: violating C aliasing rules
The bug iin `crypto_kem/hqc-rmrs-128/avx2/reed_muller.c` is found in `liboqs`—the tests suddenly fail when compiled on GCC-12, **because** for GCC this use of `tmp` is violating C aliasing rules and producing undefined behavior (thus not even showing correct compile warning):

https://github.com/PQClean/PQClean/blob/12cb25a7f04daa8c76870e6ae3ce85b4ecacd044/crypto_kem/hqc-rmrs-128/avx2/reed_muller.c#L336

It seems such indexed access to `tmp` is producing garbage. See this small discussion for details: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=106470

thomwiggers:
Closes #453


dstebila (approved):

vt-alt:
Thanks!

PR #462, opened 2022-11-01
Add VLA support (closes #452)

### LINKED ISSUE #452: Revisit VLAs?
Variable-length arrays are kind of annoying not to have. We might extend `compat.h` with a macro that implements VLAs on platforms that support them, and uses `alloca` on platforms that don't. See also for example https://github.com/sphincs/sphincsplus/pull/26. 

Of course, that might be prohibitive for any applications that don't have `alloca`. But the way we currently work around VLAs in SPHINCS+ is extremely terrible for legibility as well.

Do we want to allow VLAs, using a macro so that they will work on MSVC?

thomwiggers:


thomwiggers:
The tests here don't really mean anything as they don't use this file

dstebila (approved):

PR #465, opened 2022-11-04
Update Falcon to version included in SUPERCOP-20221028

### LINKED ISSUE #464: Malleable Falcon signatures, may need upgrade to v1.2 spec
Hi there,

The clean/avx2 implementations of Falcon-512 and Falcon-1024 seem to be unaligned with some changes introduced in v1.2 of the spec.

In particular, the implementation of the signature verification algorithm currently in PQClean accepts mauled signatures due to how `comp_decode` in `codec.c` is implemented. I attached an example program for ease of inspection.

Following Section 3.11.2 and Algorithms 17-18 in the Falcon spec v1.2, I believe that in order to enforce a unique encoding of compressed signatures, the decode implementation should not accept 100000001 ("-0") as a valid encoding of 0, and should check that signatures are padded using zeroes.

I have also raised the [same issue](https://github.com/open-quantum-safe/liboqs/issues/1315) in LibOQS for completeness.

Kind Regards,
Giacomo Fenzi

Attached bug: [bug.c](https://gist.github.com/WizardOfMenlo/59e6e0457f94bf5caf13909f319007de)

thomwiggers:
Updates Falcon.

Hopefully closes #464?


thomwiggers:
This doesn't fix the problem in #464, and only moves code from `.c` files to header files. It turns out Falcon in SUPERCOP hasn't changed since @jschanck's last PR.

PR #463, opened 2022-11-01
Add crypto_declassify.h shim

thomwiggers:
to easier support TIMECOP-supporting schemes like McEliece

PR #460, opened 2022-11-01
Update Classic McEliece

thomwiggers:
This updates classic mceliece to the SUPERCOP-20221025 version. I'm generating this version from the SUPERCOP source code, and apply patches that are listed in the repository below.

Some of the found code flaws have been reported to the submission team.

The repository where this codes is coming from can be found at
https://github.com/thomwiggers/mceliece-clean in the update-mceliece branch.

Closes #459 

## Additional content
* I'm adding VLA support through a header (#462)
* I'm adding a `crypto_declassify.h` dummy header for TIMECOP support. (#463)
* I'm dropping the `vec` and `sse` implementations of Classic McEliece.

#### Manually checked properties
* [x] Generated Github workflow (run ``.github/workflows/generate_workflows.py``) (new schemes)
* [x] No stringification macros
* [ ] Output-parameter pointers in functions are on the left
   * This would require large amounts of changes that I'm not sure are worth it
* [ ] Negative return values on failure of API functions (within restrictions of FO transform).
* [x] variable declarations at the beginning (except in `for (size_t i=...`)
* Optional:
  * [ ] All integer types are of fixed size, using `stdint.h` types (including `uint8_t` instead of `unsigned char`)
  * [ ] Integers used for indexing are of size `size_t`

PR #469, opened 2022-12-02
Update Classic McEliece

### LINKED ISSUE #459: Upgrade Classic McEliece to NIST round 4 specification
Hi! Are there any plans to upgrade the Classic McEliece KEM to the [round 4 specification](https://classic.mceliece.org/nist.html)? I suppose it will be done. But I could not find any issue or PR on it. So I figured I could create this tracking issue.

thomwiggers:
Redo of #460. Apparently you can't reopen a PR after force-pushing to the branch. (I was trying to clean the cache of changed files...)

This updates classic mceliece to the SUPERCOP-20221025 version. I'm generating this version from the SUPERCOP source code, and apply patches that are listed in the repository below.

Some of the found code flaws have been reported to the submission team.

The repository where this codes is coming from can be found at
https://github.com/thomwiggers/mceliece-clean in the update-mceliece branch.

Closes #459 

## Additional content
* I'm adding VLA support through a header (#462)
* I'm adding a `crypto_declassify.h` dummy header for TIMECOP support. (#463)
* I'm dropping the `vec` and `sse` implementations of Classic McEliece.

* [x] Generated Github workflow (run ``.github/workflows/generate_workflows.py``) (new schemes)
   * This would require large amounts of changes that I'm not sure are worth it
* Optional:


thomwiggers:
For weird reasons, the other implementations have not been picked up by CI...?

dstebila:
I've read the first three commits, now on to the big commit. I see in the commit message it's based on patches applied to the source pulled from supercop.  Would it be better to read the patches -- if so where are they?  Or what strategy do you recommend for reviewing?

thomwiggers:
This folder contains all the patches https://github.com/thomwiggers/mceliece-clean/tree/update-mceliece/patches

Clang-tidy will also patch some things afterwards, but the major stuff should be in the above.

praveksharma:
I can't copy code from d36d06981f06197044b07926ba3152c81ad2d848 to liboqs. The copy script complains about missing implementations -- d36d06981f06197044b07926ba3152c81ad2d848 only has `avx2` and `clean` implementations whereas 245c95cd1ec326f8f38e26cb17a28832701ba17b (the commit which liboqs currently relies on) has `avx`, `clean`, `vec`, and `sse` implementations.  

I'm new to both PQClean and liboqs so I'm not sure what to make of this. Does this PR require to include these other implementations before it can be merged?

PR #478, opened 2023-02-14
Fix mistake in keygen patch for Falcon

### LINKED ISSUE #477: Falcon produces invalid signatures
It seems I've introduced a bug in the Falcon update (#476) while patching some PQClean complaints. I'm currently tracking them down.

thomwiggers:
See #325 for the original patch set.


dstebila (approved):

dstebila:
If I read this correctly, it's restoring lines of code that were present in earlier versions of Falcon keygen that were mistakenly removed during the most recent update.  If so then it looks fine, but I can't attest to much beyond that.

PR #479, opened 2023-02-19
Fix Falcon signature computation

### LINKED ISSUE #477: Falcon produces invalid signatures

thomwiggers:
Falcon computes the signature in a loop, retrying if the signature is
too long. The Falcon-1024 limit seems a bit stricter than the Falcon-512
limit, which seemingly hits the recomputation more often.

If it would try to try again for a signature, the implementation would
read from the `hm` buffer again. However, `hm` was specified as a union
with `sig` and the previously attempted signature would be in its place.

In a previous version of Falcon, this recomputation would never happen,
so the case was never hit. Additionally, with the larger signature
lengths present in the NIST reference code and previous PQClean versions
of Falcon, this case would also never be hit.

Thanks a lot to @pornin for finding out this problem.

Closes #477

### DIFF
--- a/crypto_sign/falcon-1024/META.yml
+++ b/crypto_sign/falcon-1024/META.yml
@@ -22,9 +22,9 @@ auxiliary-submitters:
   - Zhenfei Zhang
 implementations:
   - name: clean
-    version: 20230207
+    version: 20211101 with PQClean patches
   - name: avx2
-    version: 20230207
+    version: 20211101 with PQClean patches
     supported_platforms:
       - architecture: x86_64
         required_flags:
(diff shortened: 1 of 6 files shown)


crypto_sign/falcon-1024/META.yml
```diff
@@ -22,9 +22,9 @@ auxiliary-submitters:
   - Zhenfei Zhang
 implementations:
   - name: clean
-    version: 20230207
+    version: 20211101 with PQClean patches
```
dstebila, line 25: Do we know if the Falcon team has any other name for this? E.g. version 2.1 or something like that?

dstebila (approved):

crypto_sign/falcon-1024/META.yml
thomwiggers: This is the version number of the Falcon zip file from `falcon-sign.info`, which my source package has no differences with other than our own patches.

dstebila: Oh, so there are no other changes to Falcon (algorithmically) since then?  Okay then.

thomwiggers: Yeah I think the spec speaks of version 1.2?

PR #487, opened 2023-05-01
Avoid -Wstrict-prototypes warning

thomwiggers:
Avoid e.g.:

```
crypto_sign/printparams.c:8:9: error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]
int main() {
        ^
         void
1 error generated.
```

### DIFF
--- a/test/crypto_kem/nistkat.c
+++ b/test/crypto_kem/nistkat.c
@@ -38,8 +38,7 @@ static void fprintBstr(FILE *fp, const char *S, const uint8_t *A, size_t L) {
     fprintf(fp, "\n");
 }
 
-int main() {
-
+int main(void) {
     uint8_t entropy_input[48];
     uint8_t seed[48];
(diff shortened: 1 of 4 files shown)


test/crypto_kem/nistkat.c
```diff
@@ -38,7 +38,9 @@ static void fprintBstr(FILE *fp, const char *S, const uint8_t *A, size_t L) {
     fprintf(fp, "\n");
 }
 
-int main() {
+int main(int argc, char** argv) {
```
dstebila, line 41: Wouldn't the following suffice?

```suggestion
int main(void) {
```

thomwiggers: Yeah I keep being confused if that's okay or not... But we have it in a bunch of different places as well I suppose.

dstebila (approved):

PR #488, opened 2023-05-02
Licence for Dilithium and Kyber

vincentvbh:
As titled.
At the beginning of each file, you will find the license information.

* Optional:


crypto_kem/kyber1024/aarch64/api.h
```diff
@@ -1,3 +1,11 @@
+
+/*
+ * This file is licensed
+ * under Apache 2.0 (https://www.apache.org/licenses/LICENSE-2.0.html) or
```
dstebila, line 4: Some of the files are MIT, some are Apache 2, is there a reason?

crypto_kem/kyber1024/aarch64/indcpa.c
```diff
@@ -1,15 +1,50 @@
...
-#include "ntt.h"
+
+/*
+ * This file was originally licensed
+ * under Apache 2.0 (https://www.apache.org/licenses/LICENSE-2.0.html)
+ * at https://github.com/GMUCERG/PQC_NEON/blob/main/neon/kyber or
+ * public domain at https://github.com/cothan/kyber/blob/master/neon
+ *
+ * We choose
+ * CC0 1.0 Universal or the following MIT License for this file.
```
dstebila, line 9: This is also really confusing to me. Are you trying to say it's triple licensed?

crypto_kem/kyber1024/aarch64/api.h
vincentvbh: We didn't change this file. The licensing is from where we got the file.

crypto_kem/kyber1024/aarch64/indcpa.c
vincentvbh: No, we license it under CC0 and MIT. The first lines say where the unmodified file comes from.

vincentvbh: For clarification, the first lines say where the file came from before modification. We modify and license it under CC0 and MIT.

mkannwischer:
Can we merge this or do we need to change it somehow? I would not know how to do it any better.

dstebila (approved):

PR #489, opened 2023-05-04
fix(SPHINCS+): another unsigned conversion in SPHINCS+

thomwiggers:
Addressing another issue reported on M2 macos by @mouse07410.


thomwiggers:
I am not fixing `-robust` or `-haraka` instances, so #505  should probably be merged first

PR #497, opened 2023-08-12
Fix publickey output sometime out of bound 

cothan:
Sometimes the key generation has error is because the public key is not properly in the bound [0, Q).
This PR make sure the output of public key is always in the bound.


cothan:
@dstebila Can you review this PR? 

I tested this update in liboqs and it works fine.

dstebila (dismissed):

test/test_format.py
```diff
@@ -12,6 +12,8 @@
 @helpers.filtered_test
 def test_format(implementation: pqclean.Implementation):
     helpers.ensure_available('astyle')
+    result = helpers.run_subprocess(['astyle', '--version'])
+    print(result)
```
thomwiggers, line 16: I don't think we intend to commit this

test/test_preprocessor.py
```diff
@@ -20,7 +20,7 @@ def test_preprocessor(implementation: pqclean.Implementation):
                 line = line.strip()
                 if file in hfiles and i == 0 and line.startswith('#ifndef'):
                     continue
-                if line.startswith('#if') and ("crypto_sign/falcon-512/avx2/sign.c" not in file) and ("crypto_sign/falcon-1024/avx2/sign.c" not in file):
+                if line.startswith('#if') and ("crypto_sign/falcon-512/avx2/sign.c" not in file) and ("crypto_sign/falcon-1024/avx2/sign.c" not in file) and ("crypto_sign\\falcon-512\\avx2\\sign.c" not in file) and ("crypto_sign\\falcon-1024\\avx2\\sign.c" not in file):
```
thomwiggers, line 23:
```suggestion
                if (line.startswith('#if') and ("crypto_sign/falcon-512/avx2/sign.c" not in file) and 
                    ("crypto_sign/falcon-1024/avx2/sign.c" not in file) and
                    ("crypto_sign\\falcon-512\\avx2\\sign.c" not in file) and
                    ("crypto_sign\\falcon-1024\\avx2\\sign.c" not in file)):
```

thomwiggers (changes requested):
Some minor nits

test/test_format.py
dstebila: I do, actually; different astyle versions lead to different formatting, so it can be useful to have that information. Unless it causes a problem elsewhere?

dstebila: Although it seems like astyle is fighting with itself on different platforms. I'm not sure how much we want to get that resolved?

cothan: I agree with @dstebila , my previous code passed all astyle checks, but this time astyle test failed in the non-commit code.

thomwiggers:
We should probably merge this and revisit all those failing tests, and fixing the astyle version, at a later point

Issue #495, opened 2023-07-20
Kyber for aarch64: incorrect Kyber768 and Kyber1024 implementation & symbol names mismatch


bhess:
The aarch64 implementations of Kyber768 and Kyber1024 seem to be incorrect and don't pass KAT tests.

The cause is in indcpa.c, which for all three Kyber variants appear to implement the same case for K=2:
https://github.com/PQClean/PQClean/blob/master/crypto_kem/kyber768/aarch64/indcpa.c
https://github.com/PQClean/PQClean/blob/master/crypto_kem/kyber1024/aarch64/indcpa.c

Another issue are symbol name mismatches between assembly code and C code, which causes linking to fail.

Example:
https://github.com/PQClean/PQClean/blob/c3abebf4ab1ff516ffa71e6337f06d898952c299/crypto_kem/kyber1024/aarch64/__asm_poly.S#L31
https://github.com/PQClean/PQClean/blob/c3abebf4ab1ff516ffa71e6337f06d898952c299/crypto_kem/kyber1024/aarch64/neon_poly.c#L134

This is also an issue in the dilithium/aarch64 code:
https://github.com/PQClean/PQClean/blob/c3abebf4ab1ff516ffa71e6337f06d898952c299/crypto_sign/dilithium2/aarch64/__asm_NTT.S#L31
https://github.com/PQClean/PQClean/blob/c3abebf4ab1ff516ffa71e6337f06d898952c299/crypto_sign/dilithium2/aarch64/ntt.h#L39

The issues were noticed while integrating the latest pqclean code to liboqs (https://github.com/open-quantum-safe/liboqs/pull/1512). The liboqs-PR contains a few temporary patches with fixes for them.

bhess:
A related question: While https://github.com/PQClean/PQClean/pull/488 added license headers to individual source files of Kyber/Dilithium/aarch64, the main [LICENSE](https://github.com/PQClean/PQClean/blob/master/crypto_kem/kyber1024/aarch64/LICENSE) files mention only CC0. Was this an oversight or intentional?

PR #501, opened 2023-08-22
Fix Dilithium Aarch64 duplicate consistency and asm

thomwiggers:
* ASM symbols were named differently than their header files (#495)
* Duplicate consistency did not work correctly due to incorrect capitalization of symbol names


thomwiggers:
yolo-merging this as I hope this will get fixed and tested by the update to the draft standard

PR #500, opened 2023-08-22
chore(formatting): Update Astyle to 3.4

thomwiggers:
Seems that Astyle has started receiving maintenance again, with several releases in 2023. This PR runs Astyle 3.4's formatter over the whole repo. Aside from a lot of empty lines that got deleted, it seems that we have a few minor improvements in indentation across the repo.

PR #511, opened 2023-09-12
Dilithium: Update to NIST Draft Standard 

mkannwischer:
* Optional:


thomwiggers:
`test_functest.py` seems to work on M2

thomwiggers:
I've merged https://github.com/PQClean/PQClean/pull/500. This means that there are a gajillion files in conflict but should also mean that we approach green CI a bit sooner.

mkannwischer:
> I've merged #500. This means that there are a gajillion files in conflict but should also mean that we approach green CI a bit sooner.

I've rebased this on top of master. So I think it's ready for review now. 
Can you test the nistkat on Apple silicon as well, please? Now it should be working fine.

thomwiggers:
Seems to work. Just checking, like the Kyber PR, this follows the reference code repository's `standard` branch, right? There are some differences I believe...

thomwiggers (dismissed):
Scrolling through this looks convincing.

mkannwischer:
> Seems to work. Just checking, like the Kyber PR, this follows the reference code repository's `standard` branch, right? There are some differences I believe...

Yes, indeed. This version is compatible with the https://github.com/pq-crystals/dilithium/tree/standard (commit https://github.com/pq-crystals/dilithium/commit/918af1a6eaedcedf9fdd8aaaca6c1fccd5a7a51f).

thomwiggers (approved):

PR #512, opened 2023-09-18
Update HQC to 2023-04-30 submission

### LINKED ISSUE #482: HQC implementation out of date/vulnerable
The current HQC implementation still contains the timing side-channel vulnerable rejection sampling, whose running time is dependent on the seed that is derived from the secret message.

https://github.com/PQClean/PQClean/blob/f1ef56679ea9b32be43619c4b4d395e7217f228b/crypto_kem/hqc-rmrs-128/avx2/vector.c#L49-L54

See [here](https://eprint.iacr.org/2021/1485.pdf) for details.

There's a new [round 4 submission](https://csrc.nist.gov/csrc/media/Projects/post-quantum-cryptography/documents/round-4/submissions/HQC-Round4.zip) that uses [Sendrier's countermeasure](https://eprint.iacr.org/2021/1631.pdf) (see `vector.c`).

Unfortunately with `gcc` the modulo reductions in the loop compile to divide instructions, which might not be constant-time.

image

`clang` successfully compiles the modulo reductions down to (I believe) Barrett reductions (which don't use `div` instructions). 

I don't know how practical it is however to exploit any leakage from this on x86, however the `clean` version which might be used to target microprocessors where division might be very leaky. 

One could manually inline the assembly to ensure that the timing behavior of the implementation doesn't depend on whether the compiler is having a good day or not :slightly_smiling_face:

SWilson4:
Update HQC to conform with the latest submission: see [https://pqc-hqc.org](https://pqc-hqc.org). This also removes the avx2 optimized version of HQC, as it is not interoperable with the latest standard and renames HQC-RMRS to HQC everywhere, as there are no longer non-RMRS versions of HQC.

I made a number of small changes to the reference code, mostly to eliminate undefined or implementation-defined behaviour. The only substantial change is the addition of a Barrett reduction routine in the code for constant-weight vector sampling. This was to address #482. Please pay special attention to this code when reviewing; it's found in vector.c.

The latest submission of HQC generates KATs using a custom PRNG instead of the one provided by NIST. I added an implementation of this PRNG in the `test` directory and modified the test files so that it would be used in the NIST KAT tests.

It might be easier to review against the reference code: [my fork of the `package-pqclean` repo](https://github.com/SWilson4/package-pqclean/tree/main/hqc) can help with this.

Fixes #482.

* [x] Generated Github workflow (run ``.github/workflows/generate_workflows.py``) (new schemes)
* Optional:


SWilson4:
The HQC tests are only failing on PPC, due to `astyle` missing: https://github.com/PQClean/PQClean/actions/runs/6252731989/job/16979928238?pr=512#step:6:4502. I've run the non-formatting tests on this branch locally using the PPC container and all the tests pass (except the preprocessor conditional checks for Kyber and Dilithium).

thomwiggers:
Yeah, we need to merge #500 after which astyle should not run on ppc anymore.

thomwiggers (changes requested):
Wait, what's happening with the NIST kat random number generators? You're touching an awful lot of meta files.

It might also be better that, if this is necessary, `nist-kat: nist` is the default when not included in a META file.

Additionally, rebasing would be great

thomwiggers:
I think this looks plausible. @mkannwischer, could you maybe also have a look at the `vector.c` barret reduction mentioned?

thomwiggers (approved):

PR #513, opened 2023-09-20
fix(tests): Fix duplicate consistency checks for SPHINCS+

thomwiggers:
The generator that I used for the duplicate consistency files for SPHINCS+ still had some assumptions on robust and haraka variants existing. This updates those files.

I'm also sneaking in a small improvement to `pqclean.py` to add the file that was not found to the exception message.


mkannwischer:
LGTM

mkannwischer (approved):

Issue #520, opened 2023-10-19
Astyle formatting differences


thomwiggers:
I've tracked down why the formatter isn't happy:

https://gitlab.com/saalen/astyle/-/issues/16#note_1610276892

PR #521, opened 2023-10-23
Stop using bad astyle versions

thomwiggers:
These two AStyle versions have a bug that breaks our formatting.

#520


thomwiggers:
doesn't solve all problems but it's a start

PR #540, opened 2024-01-24
Update README.md

aabmets:
Added reference to QuantCrypt into the README.md file. 
_**Note:**_ as per @thomwiggers request, I included a security statement into the README.md of the QuantCrypt project.


thomwiggers:
See also https://github.com/PQClean/PQClean/discussions/539

PR #541, opened 2024-01-25
Add tests for variable-length signatures

SWilson4:
This PR adds tests for variable-length signatures, as discussed in #530.

There are two tests:
1. Fill the signature buffer with random bytes before calling the signing operation. Check to see that the "tail" of the buffer (beyond the returned `siglen` is unmodified.
2. Mark the signature buffer as "undefined" and use Valgrind to check that the signing operation does not depend on the value of this buffer. This, combined with the previous test, ensures that the signing operation does not read past the end of the buffer in any meaningful way.

The commit history contains a number of reverted commits which deliberately trigger test failures. This is to demonstrate that the new tests do indeed catch undesired behaviour. If you want to double-check this, just checkout one of the "TO BE REVERTED" commits and run either
`pytest -v -n=auto -k "falcon-512 and clean" test_functest.py::test_functest` or `pytest -v -n=auto -k "falcon-512 and clean" test_valgrind.py`. I'll squash these commits at merge.

Will mark as ready for review once I see that CI has passed.

### DIFF
--- a/test/crypto_sign/functest.c
+++ b/test/crypto_sign/functest.c
@@ -5,7 +5,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#ifdef PQCLEAN_USE_VALGRIND
 #include <valgrind/memcheck.h>
+#endif
 
 #ifndef NTESTS
 #define NTESTS 5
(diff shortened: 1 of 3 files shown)


test/crypto_sign/functest.c
```diff
@@ -5,6 +5,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <valgrind/memcheck.h>
```
thomwiggers, line 8: Can we hide the Valgrind tests behind a `#define`? I think `PQCLEAN_USE_VALGRIND` or something along those lines sounds good. 

Then we can pass this define in `test_valgrind.py`.

test/crypto_sign/functest.c
```diff
@@ -124,8 +134,25 @@ static int test_sign(void) {
...
+        // Fill the sm buffer with random bytes
+        randombytes(sm_random_cmp, MLEN + CRYPTO_BYTES);
+        memcpy(sm + 8, sm_random_cmp, MLEN + CRYPTO_BYTES);
+
+        /*
+         * With this buffer marked as undefined, valgrind will detect
+         * cases where the signing code depends on the value of the tail
+         * of the buffer.
+         */
+        VALGRIND_MAKE_MEM_UNDEFINED(sm + 8, MLEN + CRYPTO_BYTES);
```
thomwiggers, line 146: It's a bit of a shame that Valgrind is required, while I have the impression that `asan` is a bit more mature nowadays (and faster). We'd ideally also want to trigger these in ASAN. Maybe we can kill two birds with one stone by instead using `malloc` to request fresh memory inside the loop. This should always be unallocated, and perhaps a bit easier for both Valgrind and ASAN to measure than the stack.

What do you think?

thomwiggers: But of course, then we can't do both checks in one go that we are doing here...

thomwiggers (changes requested):
One minor suggestion for a change, and one discussion point.

test/crypto_sign/functest.c
SWilson4: For that I believe we'd have to pass `-fsanitize=memory`, which is supported by Clang but not GCC. As far as I can tell, ASAN and UBSAN don't catch uninitialized reads. I suppose the sanitizer tests could be forced to use Clang, but apparently `-fsanitize=memory` and `-fsanitize=address` are incompatible, so it would require two passes.

test/crypto_sign/functest.c
SWilson4: Done! Also rebased on the latest `master`, so hopefully CI will be green(er).

thomwiggers:
I'm on vacation, so I'm afraid it'll be a little bit

thomwiggers (approved):

Issue #557, opened 2024-06-04
Consider renaming Kyber to ML-KEM


zugzwang:
I am testing my implementations of both Kyber round 3 and ML-KEM.

Currently, PQClean/crypto_kem/kyber{512,768,1024} implement the NIST draft, henceforth ML-KEM. It is not clear from the documentation of the [pqcrypto_kyber crate](https://docs.rs/pqcrypto-kyber/latest/pqcrypto_kyber/) nor this repository that this is the case. If you prefer to honor the original name, I would still definitely warn that the draft is implemented.

Do note that I am not trying to nitpick on names; there is a number of currently deployed projects that use e.g. Kyber768 round 3, so confusion might arise. In all honesty I spent a while debugging under the assumption that this was kyber round 3, until I realized the lack of KDF(K' || H(c)) in encaps/decaps.

aabmets:
@zugzwang If I remember it correctly, one does not just rename Kyber into ML-KEM, because NIST requested some tweaks to the original Kyber source code before it was standardized as ML-KEM, therefore it produces different output. ML-KEM would need to be included into this library explicitly by taking a look at what changes NIST specifically introduced to the algorithm and then including the source code into PQClean, retaining the original Kyber implementation, because there are dependencies by now and renaming Kyber to ML-KEM would definitely fuck up someones unittests or deployment pipelines or worse.

mkannwischer:
@aabmets @zugzwang you are both correct that Kyber round 3 and Kyber as in [FIPS203.ipd](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.ipd) have slight differences. We currently have the latter (-ish, there is some slight differences between https://github.com/pq-crystals/kyber/tree/standard and FIPS203.ipd)  in PQClean and you can still retrieve the former at https://github.com/PQClean/PQClean/releases/tag/round3.

We don't know yet what the final FIPS203 will look like and NIST is still considering some changes that would affect testvectors. I think renaming once the final standard is out and considering everything before that as different versions of Kyber is the safest option to not cause any issues downstream.

aabmets:
@mkannwischer If PQClean is going to rename Kyber to ML-KEM and Dilithium to ML-DSS, it would be pertinent and wise to put out a bulletin notifying of the change well before the change is implemented, so that third parties would have the opportunity to mitigate any issues that might arise from breaking compatibility.

mkannwischer:
[...]

PR #564, opened 2024-08-21
Update to FIPS203

### LINKED ISSUE #560: Update ML-KEM/Kyber to be compliant with FIPS203
https://csrc.nist.gov/pubs/fips/203/final

mkannwischer:
Resolves #560


thomwiggers:
Hrm, this fails on MacOS because the Aarch64 implementation is not updated.

Also, what do you think about renaming to ML-KEM? That would resolve #557

thomwiggers:
you work quick 😄

mkannwischer:
The upstream changes made to kem.c seem unnecessary - I'll wait for those to be reverted. 

I'm okay with changing the folder names to `ml-kem-{512,768,1024}` and upating the scheme name in the `META.yml` which means we also need to update the namespacing. Anything else you'd like to change?

I've also just ported the KyberSlash and clang branch vulnerability fixed to the AArch64 implementation. I'll first make sure that those actually work as expected.

thomwiggers (approved):

mkannwischer:
The changes to kem.c have been reverted upstream and it also got merged to main in https://github.com/pq-crystals/kyber/commit/10b478fc3cc4ff6215eb0b6a11bd758bf0929cbd. I pulled in that version now. 

Upstream now has the following code in `cmov` in https://github.com/pq-crystals/kyber/blob/main/ref/verify.c
```
#if defined(__GNUC__) || defined(__clang__)
  // Prevent the compiler from
  //    1) inferring that b is 0/1-valued, and
  //    2) handling the two cases with a branch.
  // This is not necessary when verify.c and kem.c are separate translation
  // units, but we expect that downstream consumers will copy this code and/or
  // change how it is built.
  __asm__("" : "+r"(b) : /* no inputs */);
#endif
```

PQClean does not allow pre-processor conditionals and `__asm__` won't work on Windows. I've removed this code for now.
(Actually our tests using pycparser also don't work with this piece of code). 
Any other ideas?

thomwiggers:
We actually have a workaround now: `common/compat.h` can contain fancy constructions as long as `test/common/compat.h` contains a non-fancy construction. It's a bit annoying that the `asm` macros are not compatible with `MSVC` though, but that's perhaps forgivable for now.

thomwiggers:
So the way to do this would be to define a `PQCLEAN_HOW_DO_WE_CALL_THIS_FENCE()` macro.

franslundberg:
Looking forward to this merge. A release after that would be appreciated. Many like me know, looking for FIPS203 implementations. Thanks for your work.

mkannwischer:
Currently we are still missing some input validation that is mandated by FIPS203, but I'm okay with merging this now and doing a follow-up PR once the reference implementation is updated. 
@thomwiggers what do you think?

thomwiggers:
Since we're not targeting FIPS certification anyway and it will be compatible (right?) I think we can merge.

thomwiggers:
Let's also pull in the `cmov` change in that follow-up PR (so I guess this code may be vulnerable to [Clangover](https://pqshield.com/pqshield-plugs-timing-leaks-in-kyber-ml-kem-to-improve-pqc-implementation-maturity/) again).

thomwiggers:
I created #566, if that's merged we can easily integrate it here and merge.

PR #566, opened 2024-09-09
Add prevent branching __asm__ macro trick

thomwiggers:
This macro can be used to prevent branching in CMOVs and such, to prevent attacks like [Clangover](https://pqshield.com/pqshield-plugs-timing-leaks-in-kyber-ml-kem-to-improve-pqc-implementation-maturity/). Unfortunately, I don't know how to do this on MSVC or in other compilers.

### DIFF
--- a/common/compat.h
+++ b/common/compat.h
@@ -54,4 +54,20 @@ UNALIGNED_VECTOR_POLYFILL_GCC
 # define PQCLEAN_VLA(__t,__x,__s) __t __x[__s]
 #endif
 
+/*********************************
+ * Prevent branching on variable *
+ *********************************/
+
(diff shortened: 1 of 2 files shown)


thomwiggers:
For reference, it appears that inline assembly is [**completely unsupported** on MSVC on platforms that are not 32-bit x86](https://learn.microsoft.com/en-us/cpp/assembler/inline/inline-assembler?view=msvc-170).

mkannwischer (approved):
Thanks this looks good to me and the CI errors are unrelated (due to a new clang version) - I'm trying to fix them separately (#568) - let's merge that first and then rebase.

PR #573, opened 2024-10-14
Add new NIST signature APIs (optionally)

mkannwischer:
I'm currently trying to add ML-DSA (Dilithium, FIPS204) to PQClean. This is the first preparatory PR.

As NIST has changed the signature APIs for the final [FIPS204](https://csrc.nist.gov/pubs/fips/204/final) and [FIPS205](https://csrc.nist.gov/pubs/fips/205/final), we additionally need to support the new APIs. Implementations may use the new APIs by defining `#define PQCLEAN_USE_SIGN_CTX` in `api.h`. 

The updated APIs following https://github.com/pq-crystals/dilithium are:

```c
int PQCLEAN_SCHEME_IMPL_crypto_sign_signature(uint8_t *sig, size_t *siglen,
        const uint8_t *m, size_t mlen,
        const uint8_t *ctx, size_t ctxlen,
        const uint8_t *sk);

int PQCLEAN_SCHEME_IMPL_crypto_sign(uint8_t *sm, size_t *smlen,
        const uint8_t *m, size_t mlen,
        const uint8_t *ctx, size_t ctxlen,
        const uint8_t *sk);

int PQCLEAN_SCHEME_IMPL_crypto_sign_verify(const uint8_t *sig, size_t siglen,
        const uint8_t *m, size_t mlen,
        const uint8_t *ctx, size_t ctxlen,
        const uint8_t *pk);

int PQCLEAN_SCHEME_IMPL_crypto_sign_open(uint8_t *m, size_t *mlen,
        const uint8_t *sm, size_t smlen,
        const uint8_t *ctx, size_t ctxlen,
        const uint8_t *pk);
```

I've switched Diltihium to the new API, but have not updated it to the new version yet.

### DIFF
--- a/test/crypto_sign/functest.c
+++ b/test/crypto_sign/functest.c
@@ -28,6 +28,10 @@
 
 #define MLEN 1024
 
+#ifdef PQCLEAN_USE_SIGN_CTX
+#define PQCLEAN_CTXLEN 14
+#endif
+
 const uint8_t canary[8] = {
     0x01, 0x23, 0x45, 0x67, 0x89, 0xAB, 0xCD, 0xEF
 };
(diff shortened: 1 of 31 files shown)


test/crypto_sign/functest.c
```diff
@@ -28,6 +28,10 @@
 
 #define MLEN 1024
 
+#ifdef PQCLEAN_USE_SIGN_CTX
+#define PQCLEAN_CTXLEN 14
```
thomwiggers, line 32: Are there any limits on this context? Would it make sense to set it to a silly number like 10000, to catch out assumptions?

test/crypto_sign/functest.c
```diff
@@ -182,8 +199,12 @@ static int test_sign(void) {
 
         // By relying on m == sm we prevent having to allocate CRYPTO_BYTES
         // twice
-        if ((returncode =
-                    crypto_sign_open(sm + 8, &mlen, sm + 8, smlen, pk + 8)) != 0) {
+        #ifdef PQCLEAN_USE_SIGN_CTX
+        returncode = crypto_sign_open(sm + 8, &mlen, sm + 8, smlen, ctx + 8, PQCLEAN_CTXLEN, pk + 8);
```
thomwiggers, line 203: We should also add a test in which the context is not set/set differently, and verification fails.

PR #579, opened 2024-12-13
Add SPDX headers and copyright notices at the beginning of files.

adrien-n:
In downstream projects, files may be copied and moved around. This makes it difficult to track the copyrights and licenses without having them stored in each file.

I've tried my best to be correct but I had to retrace changes over time, find origins again and make assumptions.

I've limited myself to Falcon for two reasons. First, at the moment, my ultimate goal is Debian packaging for liboqs and it is one of the two relevant algorithms in pqclean for liboqs. Second, it's already quite a lot of changes and things to potentially discuss. Coverage can be improved later on.

BTW, sphincsplus in pqclean is also relevant for liboqs packaging but I've elected to first do changes in sphincs/sphincplus through sphincs/sphincsplus#64 (even if the two repositories have diverged).

FInally, I'm adding mentions of NIST in copyrights for the reasons I menetion in the sphincsplus PR: it's what NIST is now using in https://csrc.nist.gov/Projects/pqc-dig-sig/standardization/example-files and it's more sensible than the previous mention of a single person.

### DIFF
--- a/crypto_sign/falcon-1024/aarch64/Makefile
+++ b/crypto_sign/falcon-1024/aarch64/Makefile
@@ -1,3 +1,7 @@
+# SPDX-License-Identifier: MIT
+# Copyright: Falcon Project
+# Copyright: PQClean Project
+
 # This Makefile can be used with GNU Make or BSD Make
 
 LIB=libfalcon-1024_aarch64.a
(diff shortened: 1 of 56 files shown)


adrien-n:
AFAICT, the test failures are from `test_format.py`. It's not a surprise since I hadn't run astyle. I also have some longer lines so that tools such as licensecheck or scancode could properly extract data but I'll look at how to improve that. Probably not a blocker for the time being though.

dstebila:
In some of the .c/.h files, you've added an SPDX header, and in others you haven't. Is there a reason to do it in some but not others?

adrien-n:
I just updated the PR after adding the missing SPDX header lines and I also removed a bunch of trailing whitespace in comments which were leading to failed CI tests (hence why CI is now passing).

dstebila (approved):

PR #582, opened 2025-01-10
Fix type of steps buffer in sphincs merkle.c

rpls:
The `leaf_info_*` structs specify a `uint32_t` pointer, but the `merkle_sign` methods allocate a `unsigned` buffer. On most platforms this is effectively the same thing (i.e., a 4-byte unsigned integer), but at least on ARMv7 `uint32_t` is defined as `unsigned long`, which is also 4 bytes but the compiler will complain.


mkannwischer (approved):
LGTM. 
The CI failures are an unrelated issue with the aarch64 implementation on macos.

mkannwischer:
Apologies @rpls, I merged this too early. There are more places in the code base where this leads to issues. Let's do a follow-up PR.

PR #587, opened 2025-03-03
ML-DSA requires SHA3 support on Aarch64

thomwiggers:
Add it to the META.yml flags


crypto_sign/ml-dsa-65/META.yml
```diff
@@ -38,4 +38,5 @@ implementations:
               - Linux
               - Darwin
           required_flags:
-              - asimd
\ No newline at end of file
+              - asimd
+              -sha3
```
mkannwischer, line 42: please give me some space.

mkannwischer (approved):

PR #588, opened 2025-03-05
Embed ___GNUC_PREREQ macro as some toolchains lack features.h

icota:
`gcc-arm-none-eabi` toolchain does not ship a `features.h`. This PR does not include it but instead embeds the version checking logic from the `__GNUC_PREREQ` macro. I believe this was the only thing `features.h` was used for.


thomwiggers:
thanks

Issue #576, opened 2024-11-02
ML-DSA does not provide `*_crypto_sign_signature` and `*_crypto_sign_verify` functions


tniessen:
Unlike all other digital signature algorithms, the ML-DSA implementations do not provide `PQCLEAN_MLDSA*_CLEAN_crypto_sign_signature` and `PQCLEAN_MLDSA*_CLEAN_crypto_sign_verify` as function symbols, which are supposed to be part of the API according to the `README.md` file:

https://github.com/PQClean/PQClean/blob/c3e6861fbb0a0b2721d2599c0a68430061414f18/README.md?plain=1#L137-L150

Instead, these implementations provide macros with the same names:

https://github.com/PQClean/PQClean/blob/c3e6861fbb0a0b2721d2599c0a68430061414f18/crypto_sign/ml-dsa-44/clean/api.h#L34-L37

While this works for calling the pseudo-functions directly, it makes it impossible to create function pointers to them. However, the ability to create function pointers is essential for consumers of PQClean that wish to provide a consistent API to end users, such as [node-pqclean](https://github.com/tniessen/node-pqclean) which has to implement [wrapper functions as a workaround](https://github.com/tniessen/node-pqclean/commit/1446af5374bff1150df46ac4f7efedc13fe17987).

It also makes C++ metaprogramming more difficult because the macro names cannot be passed to C++ templates in a consistent way.

It would be great if these could be proper function symbols instead.

mkannwischer:
Okay, I can adapt that this week. There have been some updates in https://github.com/pq-crystals/dilithium anyway that I want to integrate into PQClean.

tniessen:
Thank you @mkannwischer, that would be great!

mkannwischer:
Apologies - I did not find the time to do this yet. Feel free to open a PR.

tniessen:
@mkannwischer Is it fine to send a PR against this repository, or should a PR target some upstream repository?

mkannwischer:
> [@mkannwischer](https://github.com/mkannwischer) Is it fine to send a PR against this repository, or should a PR target some upstream repository?

I think the problem here is specific to the PQClean integration, so a PR into PQClean is fine.
For other changes, upstream PRs would be prefered.

tniessen:
@mkannwischer Thank you. I gave it a try in https://github.com/PQClean/PQClean/pull/591.

PR #591, opened 2025-04-07
Implement missing function symbols for ML-DSA

tniessen:
This PR contains two separate commits:

    Implement missing function symbols for ML-DSA

    ML-DSA currently does not implement some required parts of the API as
    function symbols and instead only provides function-like macros. This
    patch replaces the macros with functions so that users can construct
    function pointers to them.

&#x200B;

    Add test_required_symbols

    This test ensures that the common API functions are exported symbols, as
    opposed to being compile-time macros or so. It does not currently cover
    the entire API surface but would nevertheless have caught an issue in
    the past.

The existing code style across the different ML-DSA implementations is not entirely consistent, and while I tried to maintain the existing style in those source files, please let me know if you have other preferences.

Fixes: #576


thomwiggers:
AStyle is a recurring pain. Unfortunately, there doesn't seem to be any reasonable, stable, and decent formatter for C code...

thomwiggers:
I clicked the "copilot code review" button for entertainment purposes, please don't take it too seriously.

thomwiggers (approved):
Looks great and the new test is hugely helpful.

PR #595, opened 2025-05-20
Fixing memory leak in MLKEM 768 - aarch64

ghost:
Calling PQClean memory of a state in function gen_matrix() for ML-KEM 768  aarch64. 
This bug results in heap memory being drained.


thomwiggers (approved):

thomwiggers:
Weird that this was not found by the CI tests that run valgrind or asan.

mkannwischer:
> Calling PQClean memory of a state in function gen_matrix() for ML-KEM 768 aarch64. This bug results in heap memory being drained.

Thanks for finding and fixing this bug!

Maybe you do want to consider switching to a more actively maintained implementation: https://github.com/pq-code-package/mlkem-native. The bug you found isn't present in that implementation. It has the same API, so it should be easy to switch.

dstebila (approved):

thomwiggers:
Please note that we are currently planning to archive the PQClean repository. The current schedule has us marking PQClean as read-only in July 2026. Please refer to issue #604 for more information and if you have any feedback.

PR #603, opened 2026-01-08
README: Add deprecation warning

mkannwischer:
* Optional:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,15 @@
 # PQClean
 
+> [!WARNING]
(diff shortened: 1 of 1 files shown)


thomwiggers (changes requested):

dstebila (dismissed):

mkannwischer:
Thanks for the suggestions @thomwiggers. Updated. Please take a look.

thomwiggers (approved):

thomwiggers (approved):

